PR #2, opened 2016-05-09
Add basic watch for ingress updates

jsravn:
Depends on #1


ingress/controller.go
```diff
@@ -0,0 +1,98 @@
...
+	log.Infof("Controller has stopped")
+}
+
+func (c *controller) watchForUpdates(watcher k8s.Watcher) {
+	for {
+		select {
+		case <-c.stop:
+			return
+		case update := <-watcher.Updates():
+			if val, ok := update.(k8s.Ingress); ok {
```
chbatey: Can we just type the channel rather than have a generic watcher? Having maintained unyped actor systems this scares me :)

jsravn: The alternative will be duplicating the watcher interface for every type we want to support. Also means we need to change the k8s client interface whenever we want to add a new type. Or if we want to send errors for example. I was trying to make a generic watcher interface we could use instead. But I definitely see your point. It's not clear the right choice to me yet.

jsravn:
@chbatey can I merge this? then you can open another PR for your stubs.

PR #4, opened 2016-05-13
Add fake k8s client and loadbalancer to allow binary to start.

chbatey:
SIGTERM handle to shutdown gracefully.


jsravn:
LGTM - assuming we are going to replace it w/ real code. (I've already done a bit of the client in another PR)

PR #5, opened 2016-05-13
Add retrieval of ingresses from k8s apiserver

jsravn:
This PR adds real retrieval of ingresses from the apiserver. It also wires it up into the `feed-ingress` binary.

Also adds some new flags:

```
$ feed-ingress -h                                                                                        (k8s-client) 
Usage of feed-ingress:
  -apiserver string
    	Kubernetes API server URL (default "https://10.254.0.1:443")
  -cacertfile string
    	file containing kubernetes ca certificate (default "/run/secrets/kubernetes.io/serviceaccount/ca.crt")
  -debug
    	enable debug logging
  -tokenfile string
    	file containing kubernetes client authentication token (default "/run/secrets/kubernetes.io/serviceaccount/token")
```

Debug was added for some low level messages I added to check things were working against a real apiserver.

### DIFF
--- a/cmd/feed-ingress/main.go
+++ b/cmd/feed-ingress/main.go
@@ -22,7 +22,7 @@ var (
 
 func init() {
 	const (
-		defaultAPIServer  = "https://10.254.0.1:443"
+		defaultAPIServer  = "https://kubernetes:443"
 		defaultCaCertFile = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 		defaultTokenFile  = "/run/secrets/kubernetes.io/serviceaccount/token"
 	)
--- a/cmd/feed-ingress/main.go
+++ b/cmd/feed-ingress/main.go
@@ -78,7 +78,7 @@ func createK8sClient() k8s.Client {
 func readFile(path string) []byte {
 	data, err := ioutil.ReadFile(path)
 	if err != nil {
-		log.Errorf("Unable to read %s: %v", caCertFile, err)
+		log.Errorf("Unable to read %s: %v", path, err)
 		os.Exit(-1)
 	}
 	return data
--- a/cmd/feed-ingress/main.go
+++ b/cmd/feed-ingress/main.go
@@ -4,25 +4,82 @@ import (
 	"os"
 	"os/signal"
 	//
+	"flag"
+
+	"io/ioutil"
+
 	log "github.com/Sirupsen/logrus"
 	"github.com/sky-uk/feed/ingress"
 	"github.com/sky-uk/feed/k8s"
 )
 
+var (
+	apiServer  string
+	caCertFile string
+	tokenFile  string
+	debug      bool
+)
+
+func init() {
+	const (
+		defaultAPIServer  = "https://10.254.0.1:443"
+		defaultCaCertFile = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+		defaultTokenFile  = "/run/secrets/kubernetes.io/serviceaccount/token"
+	)
+
+	flag.StringVar(&apiServer, "apiserver", defaultAPIServer, "Kubernetes API server URL")
+	flag.StringVar(&caCertFile, "cacertfile", defaultCaCertFile, "file containing kubernetes ca certificate")
+	flag.StringVar(&tokenFile, "tokenfile", defaultTokenFile, "file containing kubernetes client authentication token")
+	flag.BoolVar(&debug, "debug", false, "enable debug logging")
+}
+
 func main() {
+	flag.Parse()
+
+	configureLogging()
+
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt)
 
 	lb := ingress.NewLB()
-	client := k8s.NewClient()
-	controller := ingress.NewController(lb, client)
+	client := createK8sClient()
+
+	controller := ingress.New(lb, client)
 
 	go func() {
 		for sig := range c {
-			log.Infof("Signalled {}, shutting down.", sig)
+			log.Infof("Signalled %v, shutting down.", sig)
 			controller.Stop()
 		}
 	}()
 
 	controller.Run()
+}
+
+func configureLogging() {
+	if debug {
+		log.SetLevel(log.DebugLevel)
+	}
+}
+
+func createK8sClient() k8s.Client {
+	caCert := readFile(caCertFile)
+	token := string(readFile(tokenFile))
+
+	client, err := k8s.New(apiServer, caCert, token)
+	if err != nil {
+		log.Errorf("Unable to create Kubernetes client: %v", err)
+		os.Exit(-1)
+	}
+
+	return client
+}
+
+func readFile(path string) []byte {
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		log.Errorf("Unable to read %s: %v", caCertFile, err)
+		os.Exit(-1)
+	}
+	return data
 }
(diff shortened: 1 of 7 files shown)


jsravn:
Found some bugs in this when I tried hitting a real apiserver. Hold off a bit on reviewing.

jsravn:
Ok @chbatey please have a look, this is all working.

cmd/feed-ingress/main.go
```diff
@@ -4,25 +4,82 @@ import (
...
+var (
+	apiServer  string
+	caCertFile string
+	tokenFile  string
+	debug      bool
+)
+
+func init() {
+	const (
+		defaultAPIServer  = "https://10.254.0.1:443"
```
chbatey: I don't think we should put a default in here in case we deploy it to one cluster and it is pointing to anther. Maybe failing to start is better?

cmd/feed-ingress/main.go
```diff
@@ -4,25 +4,82 @@ import (
...
-			log.Infof("Signalled {}, shutting down.", sig)
+			log.Infof("Signalled %v, shutting down.", sig)
 			controller.Stop()
 		}
 	}()
 
 	controller.Run()
 }
+
+func configureLogging() {
```
chbatey: Small function? What is this, Java??

ingress/controller.go
```diff
@@ -45,7 +45,11 @@ func (c *controller) Run() {
 	}
 
 	// initialize with current state of ingress
-	c.updateLoadBalancer()
+	err = c.updateLoadBalancer()
```
chbatey: Line 44 starts a new go routine to kick off the watching then this updates the loadbalancer.

Do we want concurrent updates of the load balancer configuration?

I can add a lock inside the load balancer otherwise we need to make sure that the load balancer updates are atomic, but even if they atomic we could end up with a race condition where the initial state (this go routine) gets scheduled after the watch go routing that contains an update. WDYT? 

We should lock in the load balancer then I think we should updateLoadBalancer before starting the watch go routine.

k8s/client.go
```diff
@@ -16,27 +25,79 @@ type Client interface {
...
+// New creates a client for the kubernetes apiserver.
+func New(apiServerURL string, caCert []byte, token string) (Client, error) {
+	parsedURL, err := url.Parse(apiServerURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid url %s: %v", apiServerURL, err)
+	}
+	baseURL := strings.TrimSuffix(parsedURL.String(), "/")
+
+	pool := x509.NewCertPool()
+	pool.AppendCertsFromPEM(caCert)
```
chbatey: Should we check that it was successfully parsed?

k8s/client.go
```diff
@@ -16,27 +25,79 @@ type Client interface {
...
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{RootCAs: pool},
+	}
+	client := &http.Client{Transport: tr}
+
+	log.Debugf("Constructing client with url: %s, token: %s, caCert: %v",
```
chbatey: Given this is a one off initialisation log could we make this info?

k8s/client.go
```diff
@@ -16,27 +25,79 @@ type Client interface {
 	WatchIngresses(Watcher) error
 }
 
-type noopClient struct {
+type impl struct {
```
chbatey: Is it normal to call it impl?

k8s/client.go
```diff
@@ -16,27 +25,79 @@ type Client interface {
...
+			token:   token,
+			client:  client},
+		nil
 }
 
-func (client *noopClient) WatchIngresses(w Watcher) error {
-	log.Infof("Watching ingresses {}", w)
+func (i *impl) GetIngresses() ([]Ingress, error) {
+	endpoint := i.baseURL + "/apis/extensions/v1beta1/ingresses"
+	req, err := http.NewRequest("GET", endpoint, nil)
```
chbatey: This error isn't checked. The go compiler should really catch that this isn't read until it is reassigned :(

k8s/client_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+func TestRetrievesIngressesFromKubernetes(t *testing.T) {
+	assert := assert.New(t)
+
+	ingressFixture := CreateIngressesFixture()
+	ts := httptest.NewTLSServer(handleGetIngresses(t, authToken, ingressFixture))
+	defer ts.Close()
+
+	client, err := New(ts.URL, caCert, authToken)
+	assert.NoError(err)
```
chbatey: This oddly returns a boolean rather than failing the test. I tend to use `assert.Nil` on errros.

chbatey: Ah I had been using assert.Nil

I forgot that a test can be failed internally even if the return type is bool

chbatey:
A few minor comments. Assigned back to you @jsravn

cmd/feed-ingress/main.go
jsravn: You want it to point to another cluster? Not sure what you mean. 10.254.0.1 is the apiserver vip inside of a kubernetes cluster, it shouldn't be access any other cluster.

cmd/feed-ingress/main.go
jsravn: Is that good or bad? :)

Actually this used to have more in it. I can inline.

ingress/controller.go
jsravn: We should definitely mutex protect the loadbalancer if it needs it. Which it probably does, since separate goroutines will be accessing it.

Doing the initialisation after creating a watch is the only way to ensure we don't miss an update. Otherwise we may miss an update between the initial select and the subsequent subscription. I don't see any race condition here, the watcher will simply call getIngresses if there is an update.

We could also expose the resource version directly on the client, that would avoid all problems, and it wouldn't matter what order we call things. But it is leaking more kubernetes logic into the controller. And we'll have to return resource version from `getIngresses`.

k8s/client.go
jsravn: Yep, lots of golang libs I've looked at do that. It's a very common idiom in C++ too.

jsravn: We don't have to... let's decide on a convention.
We could also call it after the interface, but lowercase maybe? So 'client'.

k8s/client.go
jsravn: Yep!

cmd/feed-ingress/main.go
```diff
@@ -4,25 +4,82 @@ import (
...
+		os.Exit(-1)
+	}
+
+	return client
+}
+
+func readFile(path string) []byte {
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		log.Errorf("Unable to read %s: %v", caCertFile, err)
```
balooo: should this be `path` instead of `caCertFile`

k8s/client.go
jsravn: It's really noisy, the cert and tokens are big. Like it prints 10-15 lines of stuff. We could have a single info saying it's creating the client? But I'm not sure what that offers, no external interactions happen until it tries to do a request.

k8s/client.go
jsravn: Yup I already fixed this in my other branch. I can fix it here though - it was the main cause of my problems trying to debug why it wasn't working :).

ingress/controller.go
jsravn: Ok I'll fix this - you're right there is a race condition on the ingress selection/update.

cmd/feed-ingress/main.go
jsravn: Yup, thanks.

PR #6, opened 2016-05-16
Add nginx management and docker container

chbatey:
Depends on #4

I went with Alpine to keep the image small.

Alternative implementation would have the nginx in a different container but this way this image is self contained.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .idea
 *.iml
+*.swp
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,18 @@ go:
 install:
   - go get github.com/golang/lint/golint
   - go get golang.org/x/tools/cmd/goimports
+
+services:
+  - docker
+
+after_success:
+  - docker login -e $DOCKER_EMAIL -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
+  - export REPO=skycirrus/feed-ingress
+  - export TAG=`if [ "$TRAVIS_BRANCH" == "master" ]; then echo "latest"; else echo $TRAVIS_BRANCH ; fi`
+  - cd docker && docker build -f Dockerfile -t $REPO:$COMMIT .
+  - docker tag $REPO:$COMMIT $REPO:$TAG
+  - docker push $REPO
+
+env:
+  global:
+    - COMMIT=${TRAVIS_COMMIT::8}
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 pkgs = $(shell go list ./... | grep -v /vendor/)
 files = $(shell find . -path ./vendor -prune -o -name '*.go' -print)
+ingress_binary = $(GOPATH)/bin/feed-ingress
+template = ./ingress/nginx.tmpl
 
-all : format vet lint test build
+all : format vet lint test build copy
 
 format :
 	@echo "== formatting"
@@ -25,4 +27,9 @@ lint :
 		golint -set_exit_status $$pkg || exit 1; \
 	done;
 
-.PHONY: all format test build vet lint
+copy :
+	@echo "== coping files to docker"
+	cp $(ingress_binary) ./docker/
+	cp $(template) ./docker/
+
+.PHONY: all format test build vet lint copy
--- a/cmd/feed-ingress/main.go
+++ b/cmd/feed-ingress/main.go
@@ -14,22 +14,31 @@ import (
 )
 
 var (
-	apiServer  string
-	caCertFile string
-	tokenFile  string
-	debug      bool
+	nginxConfDir         string
+	ingressPort          int
+	nginxWorkerProcesses int
+	apiServer            string
+	caCertFile           string
+	tokenFile            string
+	debug                bool
 )
 
 func init() {
 	const (
-		defaultAPIServer  = "https://kubernetes:443"
-		defaultCaCertFile = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
-		defaultTokenFile  = "/run/secrets/kubernetes.io/serviceaccount/token"
+		defaultAPIServer    = "https://kubernetes:443"
+		defaultCaCertFile   = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+		defaultTokenFile    = "/run/secrets/kubernetes.io/serviceaccount/token"
+		defaultNginxConfDir = "."
+		defaultIngressPort  = 80
+		defaultNginxWorkers = 1
 	)
 
 	flag.StringVar(&apiServer, "apiserver", defaultAPIServer, "Kubernetes API server URL")
 	flag.StringVar(&caCertFile, "cacertfile", defaultCaCertFile, "file containing kubernetes ca certificate")
 	flag.StringVar(&tokenFile, "tokenfile", defaultTokenFile, "file containing kubernetes client authentication token")
+	flag.StringVar(&nginxConfDir, "nginx-conf-dir", defaultNginxConfDir, "directory to store nginx conf")
+	flag.IntVar(&ingressPort, "ingress-port", defaultIngressPort, "port to server ingress traffic")
+	flag.IntVar(&nginxWorkerProcesses, "nginx-workers", defaultNginxWorkers, "nginx worker processes")
 	flag.BoolVar(&debug, "debug", false, "enable debug logging")
 }
 
@@ -41,7 +50,18 @@ func main() {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt)
 
-	lb := ingress.NewLB()
+	lb := ingress.NewNginxLB(&ingress.NginxConf{
+		BinaryLocation:  "/usr/sbin/nginx",
+		ConfigDir:       nginxConfDir,
+		WorkerProcesses: nginxWorkerProcesses,
+		Port:            ingressPort,
+	}, &ingress.DefaultSignaller{})
+	err := lb.Start()
+	if err != nil {
+		log.Error("Unable to start nginx", err)
+		os.Exit(-1)
+	}
+
 	client := createK8sClient()
 
 	controller := ingress.New(lb, client)
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,22 @@
+FROM debian:jessie
+
+ENV NGINX_VERSION 1.10.0-1~jessie
+
+RUN apt-key adv --keyserver hkp://pgp.mit.edu:80 --recv-keys 573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62 \
+	&& echo "deb http://nginx.org/packages/debian/ jessie nginx" >> /etc/apt/sources.list \
+	&& apt-get update \
+	&& apt-get install --no-install-recommends --no-install-suggests -y \
+						ca-certificates \
+						nginx=${NGINX_VERSION} \
+						nginx-module-xslt \
+						nginx-module-geoip \
+						nginx-module-image-filter \
+						nginx-module-perl \
+						nginx-module-njs \
+						gettext-base \
+	&& rm -rf /var/lib/apt/lists/*
+
+COPY feed-ingress /
+COPY nginx.tmpl /
+
+ENTRYPOINT ["/feed-ingress"]
--- a/ingress/controller.go
+++ b/ingress/controller.go
@@ -97,10 +97,13 @@ func (c *controller) updateLoadBalancer() error {
 	}
 
 	log.Infof("Updating load balancer with %d entry(s)", len(entries))
-	return c.lb.Update(entries)
+	updated, err := c.lb.Update(LoadBalancerUpdate{entries})
+	log.Infof("Load balancer updated? %s", updated)
+	return err
 }
 
 func (c *controller) Stop() {
 	log.Info("Stopping controller")
+	c.lb.Stop()
 	close(c.stop)
 }
--- a/ingress/loadbalancer.go
+++ b/ingress/loadbalancer.go
@@ -1,35 +1,70 @@
 package ingress
 
 import (
-	log "github.com/Sirupsen/logrus"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"syscall"
 )
 
 // LoadBalancer that the controller will modify.
 type LoadBalancer interface {
-	Update([]LoadBalancerEntry) error
+	Start() error
+	Stop() error
+	// Update the loadbalancer configuration. Returns if the LB was required to reload
+	// its configuration
+	Update(LoadBalancerUpdate) (bool, error)
 }
 
-// LoadBalancerEntry describes the ingress for a single host, path, and service.
-type LoadBalancerEntry struct {
-	Host        string
-	Path        string
-	ServiceName string
-	ServicePort int32
+// Signaller interface around signalling the loadbalancer process
+type Signaller interface {
+	Sigquit(*os.Process) error
+	Sighup(*os.Process) error
 }
 
-type noopLoadBalancer struct {
+// DefaultSignaller sends os signals
+type DefaultSignaller struct {
 }
 
-func (lb *noopLoadBalancer) Update(entries []LoadBalancerEntry) error {
-	log.Debugf("Updating loadbalancer for %v", entries)
-	return nil
+// Sigquit sends a SIGQUIT to the process
+func (s *DefaultSignaller) Sigquit(p *os.Process) error {
+	return p.Signal(syscall.SIGQUIT)
 }
 
-// NewLB creates a new LoadBalancer
-func NewLB() LoadBalancer {
-	return &noopLoadBalancer{}
+// Sighup sends a SIGHUP to the process
+func (s *DefaultSignaller) Sighup(p *os.Process) error {
+	return p.Signal(syscall.SIGHUP)
 }
 
-func (lb *noopLoadBalancer) String() string {
-	return "[fake lb]"
+func writeFile(location string, contents []byte) (bool, error) {
+	err := ioutil.WriteFile(location, contents, 0644)
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+func diff(b1, b2 []byte) ([]byte, error) {
+	f1, err := ioutil.TempFile("", "")
+	if err != nil {
+		return nil, err
+	}
+	defer os.Remove(f1.Name())
+	defer f1.Close()
+
+	f2, err := ioutil.TempFile("", "")
+	if err != nil {
+		return nil, err
+	}
+	defer os.Remove(f2.Name())
+	defer f2.Close()
+
+	f1.Write(b1)
+	f2.Write(b2)
+
+	data, err := exec.Command("diff", "-u", f1.Name(), f2.Name()).CombinedOutput()
+	if len(data) > 0 {
+		return data, nil
+	}
+	return data, err
 }
--- a/ingress/nginx.tmpl
+++ b/ingress/nginx.tmpl
@@ -1,0 +1,47 @@
+worker_processes  1;
+daemon off;
+
+error_log  /dev/stdout warn;
+pid        {{ .Config.ConfigDir }}/nginx.pid;
+
+events {
+    worker_connections  1024;
+}
+
+http {
+    default_type text/html;
+
+    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+    '$status $body_bytes_sent "$http_referer" '
+    '"$http_user_agent" "$http_x_forwarded_for"';
+
+    access_log  /var/log/nginx/access.log  main;
+    sendfile        on;
+    keepalive_timeout  65;
+
+    {{ $port := .Config.Port }}
+    {{ range $entry := .Entries }}
+    # Start entry
+    server {
+        listen {{ $port }};
+        server_name {{ $entry.Host }};
+        location {{ $entry.Path }} {
+            proxy_pass http://{{ $entry.ServiceName }}:{{ $entry.ServicePort }};
+        }
+    }
+    # End entry
+    {{ end }}
+
+    # default server for health checks
+    server {
+        listen {{ .Config.Port }} default_server;
+            location /health {
+             access_log off;
+             return 200;
+        }
+
+        location / {
+            return 404;
+        }
+    }
+}
--- a/README.md
+++ b/README.md
@@ -27,6 +27,12 @@ It requires these tools:
 
     go get -u github.com/golang/lint/golint
     go get -u golang.org/x/tools/cmd/goimports
+    
+# Releasing
+
+Travis is configured to build the Docker image and push it to Dockerhub for each PR.
+
+For a proper release create a tag and push and Travis will push the image to Dockerhub.
 
 # Dependencies
 
--- a/docker/.dockerignore
+++ b/docker/.dockerignore
@@ -1,0 +1,1 @@
+.gitignore
--- a/docker/.gitignore
+++ b/docker/.gitignore
@@ -1,0 +1,4 @@
+feed-ingress
+nginx.tmpl
+nginx.conf
+nginx.pid
--- a/ingress/controller_test.go
+++ b/ingress/controller_test.go
@@ -19,9 +19,21 @@ type fakeLb struct {
 	mock.Mock
 }
 
-func (lb *fakeLb) Update(entries []LoadBalancerEntry) error {
-	r := lb.Called(entries)
-	return r.Error(0)
+func (lb *fakeLb) Update(update LoadBalancerUpdate) (bool, error) {
+	r := lb.Called(update)
+	return false, r.Error(0)
+}
+
+func (lb *fakeLb) Start() error {
+	return nil
+}
+
+func (lb *fakeLb) Stop() error {
+	return nil
+}
+
+func (lb *fakeLb) WaitFor() error {
+	return nil
 }
 
 func (lb *fakeLb) String() string {
@@ -113,7 +125,7 @@ func TestLoadBalancerUpdatesOnIngressUpdates(t *testing.T) {
 
 	//then
 	entries := createLbEntriesFixture()
-	lb.AssertCalled(t, "Update", []LoadBalancerEntry{})
+	lb.AssertCalled(t, "Update", LoadBalancerUpdate{[]LoadBalancerEntry{}})
 	lb.AssertCalled(t, "Update", entries)
 
 	//cleanup
@@ -158,13 +170,13 @@ func sendUpdate(watcher k8s.Watcher, value interface{}, d time.Duration) error {
 	return nil
 }
 
-func createLbEntriesFixture() []LoadBalancerEntry {
-	return []LoadBalancerEntry{LoadBalancerEntry{
+func createLbEntriesFixture() LoadBalancerUpdate {
+	return LoadBalancerUpdate{[]LoadBalancerEntry{LoadBalancerEntry{
 		Host:        ingressHost,
 		Path:        ingressPath,
 		ServiceName: ingressSvcName,
 		ServicePort: ingressSvcPort,
-	}}
+	}}}
 }
 
 const (
--- a/ingress/fake_nginx.sh
+++ b/ingress/fake_nginx.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+echo "$@"
+
+
--- a/ingress/lb_entry.go
+++ b/ingress/lb_entry.go
@@ -1,0 +1,50 @@
+package ingress
+
+import (
+	log "github.com/Sirupsen/logrus"
+)
+
+// LoadBalancerUpdate data required to update loadbalancing configuration
+type LoadBalancerUpdate struct {
+	Entries []LoadBalancerEntry
+}
+
+// LoadBalancerEntry describes the ingress for a single host, path, and service.
+type LoadBalancerEntry struct {
+	Host        string
+	Path        string
+	ServiceName string
+	ServicePort int32
+}
+
+// FilterInvalidEntries returns a slice of all the valid LoadBalancer entries
+func filterInvalidEntries(entries []LoadBalancerEntry) []LoadBalancerEntry {
+	var validEntries []LoadBalancerEntry
+
+	for _, entry := range entries {
+		if entry.ValidateEntry() {
+			validEntries = append(validEntries, entry)
+		} else {
+			log.Warnf("Removing invalid load balancer entry for service '%s' host '%s'", entry.ServiceName, entry.Host)
+		}
+	}
+
+	return validEntries
+}
+
+// ValidateEntry returns whether the given entry is valid
+func (entry LoadBalancerEntry) ValidateEntry() bool {
+	if entry.Host == "" {
+		return false
+	}
+	if entry.Path == "" {
+		return false
+	}
+	if entry.ServiceName == "" {
+		return false
+	}
+	if entry.ServicePort == 0 {
+		return false
+	}
+	return true
+}
--- a/ingress/lb_entry_test.go
+++ b/ingress/lb_entry_test.go
@@ -1,0 +1,85 @@
+package ingress
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestValidEntry(t *testing.T) {
+	assert.True(t, LoadBalancerEntry{
+		Host:        "valid",
+		Path:        "valid",
+		ServiceName: "valid",
+		ServicePort: 9090,
+	}.ValidateEntry())
+}
+
+func TestValidateHost(t *testing.T) {
+	assert.False(t, LoadBalancerEntry{
+		Host:        "",
+		Path:        "valid",
+		ServiceName: "valid",
+		ServicePort: 9090,
+	}.ValidateEntry())
+	assert.False(t, LoadBalancerEntry{
+		Path:        "valid",
+		ServiceName: "valid",
+		ServicePort: 9090,
+	}.ValidateEntry())
+}
+
+func TestValidatePath(t *testing.T) {
+	assert.False(t, LoadBalancerEntry{
+		Host:        "valid",
+		Path:        "",
+		ServiceName: "valid",
+		ServicePort: 9090,
+	}.ValidateEntry())
+	assert.False(t, LoadBalancerEntry{
+		Host:        "valid",
+		ServiceName: "valid",
+		ServicePort: 9090,
+	}.ValidateEntry())
+}
+
+func TestValidateServiceName(t *testing.T) {
+	assert.False(t, LoadBalancerEntry{
+		Host:        "valid",
+		Path:        "valid",
+		ServiceName: "",
+		ServicePort: 9090,
+	}.ValidateEntry())
+	assert.False(t, LoadBalancerEntry{
+		Host:        "valid",
+		Path:        "valid",
+		ServicePort: 9090,
+	}.ValidateEntry())
+}
+
+func TestValidatePort(t *testing.T) {
+	assert.False(t, LoadBalancerEntry{
+		Host:        "valid",
+		Path:        "valid",
+		ServiceName: "valid",
+		ServicePort: 0,
+	}.ValidateEntry())
+	assert.False(t, LoadBalancerEntry{
+		Host:        "valid",
+		Path:        "valid",
+		ServiceName: "valid",
+	}.ValidateEntry())
+}
+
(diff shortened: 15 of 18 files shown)


Makefile
```diff
@@ -1,7 +1,8 @@
 pkgs = $(shell go list ./... | grep -v /vendor/)
 files = $(shell find . -path ./vendor -prune -o -name '*.go' -print)
+ingress_binary = $(GOPATH)/bin/feed-ingress
 
-all : format vet lint test build
+all : format vet lint test build docker
```
jsravn: I'm not sure docker should be part of the default build. Doesn't it take quite a while to build?

docker/Dockerfile
```diff
@@ -0,0 +1,122 @@
...
+	# forward request and error logs to docker log collector
+	&& ln -sf /dev/stdout /var/log/nginx/access.log \
+	&& ln -sf /dev/stderr /var/log/nginx/error.log
+
+COPY nginx.conf /etc/nginx/nginx.conf
+COPY nginx.vh.default.conf /etc/nginx/conf.d/default.conf
+
+COPY feed-ingress /
+
+CMD ["/feed-ingress"]
```
jsravn: eol

Also, shouldn't we use `ENTRYPOINT` so we can pass our own flags in?

docker/nginx.conf
```diff
@@ -0,0 +1,32 @@
...
+    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+    '$status $body_bytes_sent "$http_referer" '
+    '"$http_user_agent" "$http_x_forwarded_for"';
+
+    access_log  /var/log/nginx/access.log  main;
+
+    sendfile        on;
+    #tcp_nopush     on;
+
+    keepalive_timeout  65;
```
jsravn: We'll probably want to configure this somehow. 65 will be too high as a default though as ELBs are currently set to 60.

docker/nginx.conf
```diff
@@ -0,0 +1,32 @@
+user  nginx;
+worker_processes  1;
```
jsravn: This should also be configurable at some point. It will be based on number of cores we give the nginx pods (probably more than 1, less than 8).

docker/Dockerfile
```diff
@@ -0,0 +1,122 @@
...
+			| sort -u \
+	)" \
+	&& apk add --virtual .nginx-rundeps $runDeps \
+	&& apk del .build-deps \
+	&& rm -rf /usr/src/nginx-$NGINX_VERSION \
+	&& apk add --no-cache gettext \
+	\
+	# forward request and error logs to docker log collector
+	&& ln -sf /dev/stdout /var/log/nginx/access.log \
+	&& ln -sf /dev/stderr /var/log/nginx/error.log
```
jsravn: Where did you copy the configure from for nginx? Is it from the official image? I have no idea what any of these flags do so we'll have to trust wherever you copied it from :)

docker/Dockerfile
```diff
@@ -0,0 +1,122 @@
+FROM alpine:3.3
```
jsravn: I believe alpine has issues in kubernetes due to dns resolution (doesn't respect search domains in /etc/resolv.conf). We also probably want the diagnostic tools you get in ubuntu.

jsravn:
lgtm apart from a few comments

Makefile
chbatey: It is very quick apart from the first time :)

docker/Dockerfile
chbatey: Yep

docker/Dockerfile
chbatey: Yeah there is a comment from the top to say that it was from the alpine official image

docker/Dockerfile
chbatey: Want to stick with alpine until we see issues? 

If not I don't mind changing ti

Makefile
jsravn: I think we should have a separate release target that will build the image.

docker/Dockerfile
jsravn: Well it simply won't work - it won't resolve services correctly.

chbatey:
@jsravn this is ready for review again

jsravn:
@chbatey can you squash the commits? Ideally the docker commit is separate to the nginx work.

.travis.yml
```diff
@@ -6,3 +6,18 @@ go:
...
+  - docker login -e $DOCKER_EMAIL -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
+  - export REPO=skycirrus/feed-ingress
+  - export TAG=`if [ "$TRAVIS_BRANCH" == "master" ]; then echo "latest"; else echo $TRAVIS_BRANCH ; fi`
+  - cd docker && docker build -f Dockerfile -t $REPO:$COMMIT .
+  - docker tag $REPO:$COMMIT $REPO:$TAG
+  - docker push $REPO
+
+env:
+  global:
+    - COMMIT=${TRAVIS_COMMIT::8}
```
jsravn: eol

Makefile
```diff
@@ -1,7 +1,9 @@
 pkgs = $(shell go list ./... | grep -v /vendor/)
 files = $(shell find . -path ./vendor -prune -o -name '*.go' -print)
+ingress_binary = $(GOPATH)/bin/feed-ingress
+template = ./ingress/nginx.tmpl
 
-all : format vet lint test build
+all : format vet lint test build docker
```
jsravn: I think `docker` should be part of a release target, separate from the default (which should be for the normal dev build cycle).

cmd/feed-ingress/main.go
```diff
@@ -41,7 +47,13 @@ func main() {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt)
 
-	lb := ingress.NewLB()
+	lb := ingress.NewLB("/usr/sbin/nginx", nginxConfDir, nginxPort, &ingress.DefaultSignaller{})
```
jsravn: Maybe we should call it `NewNginxLB`. My idea is if we support other load balancers (like brocade), then it will just be a new constructor.

docker/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+						nginx-module-geoip \
+						nginx-module-image-filter \
+						nginx-module-perl \
+						nginx-module-njs \
+						gettext-base \
+	&& rm -rf /var/lib/apt/lists/*
+
+# forward request and error logs to docker log collector
+RUN ln -sf /dev/stdout /var/log/nginx/access.log \
+	&& ln -sf /dev/stderr /var/log/nginx/error.log
```
jsravn: Won't this break all of our logging? It will all go to the log instead of the containers stdout. Can we just change nginx logfile locations?

docker/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+COPY feed-ingress /
+COPY nginx.tmpl /
+
+ENTRYPOINT ["/feed-ingress"]
```
jsravn: eol

ingress/controller.go
```diff
@@ -97,7 +97,8 @@ func (c *controller) updateLoadBalancer() error {
 	}
 
 	log.Infof("Updating load balancer with %d entry(s)", len(entries))
-	return c.lb.Update(entries)
+	_, err = c.lb.Update(LoadBalancerUpdate{entries})
```
jsravn: Isn't it a bit weird we ignore the return value? I guess it's used for tests only? Maybe we should log off of it in the controller at least.

ingress/loadbalancer_util.go
```diff
@@ -0,0 +1,39 @@
...
+		return false
+	}
+	if entry.ServiceName == "" {
+		return false
+	}
+	if entry.ServicePort == 0 {
+		return false
+	}
+	return true
+}
```
jsravn: Can this be a method on the LoadBalancerEntry itself? So we can do `entry.Validate()`?

ingress/loadbalancer_util.go
```diff
@@ -0,0 +1,39 @@
+package ingress
+
+import (
+	log "github.com/Sirupsen/logrus"
+)
+
+// Functions that benefit from isolated testing
```
jsravn: Maybe it would make sense to move LoadBalanceryEntry in here as well, and rename the file `lb_entry.go` or something like that. Then we keep the data structure with the functions that operate on it.

ingress/nginx.tmpl
```diff
@@ -0,0 +1,44 @@
+worker_processes  1;
```
jsravn: This needs to be configurable w/ a flag so we can specify at pod level (cpu request = worker_processes).

ingress/nginx.tmpl
```diff
@@ -0,0 +1,44 @@
...
+daemon off;
+
+error_log  /dev/stdout warn;
+pid        {{ .WorkingDir }}/nginx.pid;
+
+events {
+ worker_connections  1024;
+}
+
```
jsravn: Can we make formatting consistent and use 2 spaces for indent?

ingress/nginx.tmpl
```diff
@@ -0,0 +1,44 @@
+worker_processes  1;
+daemon off;
+
+error_log  /dev/stdout warn;
+pid        {{ .WorkingDir }}/nginx.pid;
+
+events {
+ worker_connections  1024;
```
jsravn: This also should be configurable - it's not clear yet what the ideal is, we'll have to capacity plan it based on expected number of persistent connections. It also depends on the number of worker processes. Although, maybe 1024 is good enough...? I'm not sure what happens if this value gets exceeded.

ingress/nginx.tmpl
```diff
@@ -0,0 +1,44 @@
+worker_processes  1;
+daemon off;
+
+error_log  /dev/stdout warn;
+pid        {{ .WorkingDir }}/nginx.pid;
+
```
jsravn: Need to increase `worker_rlimit_nofile` or we'll run out of file descriptors fast under load. I think default ulimit is 1024.

See: http://stackoverflow.com/questions/7325211/tuning-nginx-worker-process-to-obtain-100k-hits-per-min

docker/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+						ca-certificates \
+						nginx=${NGINX_VERSION} \
+						nginx-module-xslt \
+						nginx-module-geoip \
+						nginx-module-image-filter \
+						nginx-module-perl \
+						nginx-module-njs \
+						gettext-base \
+	&& rm -rf /var/lib/apt/lists/*
+
```
jsravn: Don't we want to run as nginx user? I think running the proxy as root is dangerous since it will be our main external entry point.

ingress/nginx.tmpl
```diff
@@ -0,0 +1,44 @@
...
+events {
+ worker_connections  1024;
+}
+
+http {
+ default_type  application/octet-stream;
```
jsravn: Why default to octet-stream?

ingress/nginx.tmpl
jsravn: We should also use epoll and enable multi_accept.

See https://www.linode.com/docs/websites/nginx/configure-nginx-for-optimized-performance

ingress/nginx.tmpl
```diff
@@ -0,0 +1,44 @@
...
+
+ log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+ '$status $body_bytes_sent "$http_referer" '
+ '"$http_user_agent" "$http_x_forwarded_for"';
+
+ access_log  /var/log/nginx/access.log  main;
+ sendfile        on;
+ keepalive_timeout  65;
```
jsravn: Also needs to be configurable - it depends on the external setup.

ingress/nginx.tmpl
```diff
@@ -0,0 +1,44 @@
...
+ server {
+   listen {{ $port }};
+   server_name {{ $entry.Host }};
+   location {{ $entry.Path }} {
+     proxy_pass http://{{ $entry.ServiceName }}:{{ $entry.ServicePort }};
+   }
+ }
+ # End entry
+ {{ end }}
+
```
jsravn: Do we have a default 404 ?

ingress/nginx.tmpl
```diff
@@ -0,0 +1,44 @@
...
+ # default server for health checks
+ server {
+   listen {{ .Port }} default_server;
+   location /health {
+     access_log off;
+     return 200;
+   }
```
jsravn: I think the health check should be on a separate port on localhost only. It just needs to be accessible from kubelet.

ingress/nginx.tmpl
```diff
@@ -0,0 +1,44 @@
...
+ keepalive_timeout  65;
+
+ {{ $port := .Port }}
+ {{ range $entry := .Entries }}
+ # Start entry
+ server {
+   listen {{ $port }};
+   server_name {{ $entry.Host }};
+   location {{ $entry.Path }} {
+     proxy_pass http://{{ $entry.ServiceName }}:{{ $entry.ServicePort }};
```
jsravn: We should pass in the client ip the ELBs provide us - this will be required by app teams I think, and will be important for debugging issues when external things start connecting. Can get this from the x-forward-for header from the ELB I believe.

ingress/nginx.tmpl
```diff
@@ -0,0 +1,44 @@
...
+   }
```
jsravn: Do you know if we need to use `upstreams` to pool connections to downstream? I'm not sure if nginx maintains a keepalive to a backend for every client keepalive.

ingress/nginx.tmpl
```diff
@@ -0,0 +1,44 @@
...
+ # default server for health checks
+ server {
+   listen {{ .Port }} default_server;
+   location /health {
+     access_log off;
+     return 200;
+   }
+ }
+
+}
```
jsravn: Few more things:
- Didn't we get an error message about the hash bucket size being too small? I think the default is too small for the xeon cpus we use on AWS. But not sure.
- I'd rather start with an optimized config from online (such as the one in kubernetes contrib), rather than sticking with nginx defaults - which afaik are not great out of the box.

ingress/loadbalancer.go
```diff
@@ -2,11 +2,31 @@ package ingress
```
jsravn: I think all nginx specific stuff should go into its own `nginx.go` file. Then if we add other load balancers, it's just a new file like `brocade.go`. The api interfaces should remain in here. What do you think?

ingress/loadbalancer.go
```diff
@@ -2,11 +2,31 @@ package ingress
...
 // LoadBalancer that the controller will modify.
 type LoadBalancer interface {
-	Update([]LoadBalancerEntry) error
+	Start() error
+	Stop() error
+	Update(LoadBalancerUpdate) (bool, error)
+}
+
+// Signaller interface around signalling the loadbalancer process
+type Signaller interface {
```
jsravn: Does it need to be a public interface?

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+
+// Sighup sends a SIGHUP to the process
+func (s *DefaultSignaller) Sighup(p *os.Process) error {
+	return p.Signal(syscall.SIGHUP)
 }
 
 // NewLB creates a new LoadBalancer
-func NewLB() LoadBalancer {
-	return &noopLoadBalancer{}
+func NewLB(nginxLocation, nginxConfDir string, port int, signaller Signaller) LoadBalancer {
```
jsravn: I think the signaller can be created internally - no need to require the controller to create it.

ingress/loadbalancer.go
```diff
@@ -2,11 +2,31 @@ package ingress
...
+	"text/template"
 )
 
 // LoadBalancer that the controller will modify.
 type LoadBalancer interface {
-	Update([]LoadBalancerEntry) error
+	Start() error
+	Stop() error
+	Update(LoadBalancerUpdate) (bool, error)
+}
```
jsravn: Add comments for each public interface method. I thought gofmt required this actually - maybe not.

ingress/loadbalancer.go
```diff
@@ -2,11 +2,31 @@ package ingress
...
+	"syscall"
+	"text/template"
 )
 
 // LoadBalancer that the controller will modify.
 type LoadBalancer interface {
-	Update([]LoadBalancerEntry) error
+	Start() error
+	Stop() error
+	Update(LoadBalancerUpdate) (bool, error)
```
jsravn: Can we add a comment for Update to indicate what the bool means.

ingress/loadbalancer.go
```diff
@@ -2,11 +2,31 @@ package ingress
...
+	"os"
+	"os/exec"
+	"syscall"
+	"text/template"
 )
 
 // LoadBalancer that the controller will modify.
 type LoadBalancer interface {
-	Update([]LoadBalancerEntry) error
+	Start() error
```
jsravn: Worth commenting that is async?

cmd/feed-ingress/main.go
```diff
@@ -41,7 +47,13 @@ func main() {
...
+	err := lb.Start()
+	if err != nil {
+		log.Error("Unable to start nginx", err)
+		os.Exit(-1)
+	}
+
```
jsravn: It doesn't look like `lb.Stop()` is ever called. It should happen when controller is stopped.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+	return &nginxLoadBalancer{
+		binary:    nginxLocation,
+		confDir:   nginxConfDir,
+		port:      port,
+		signaller: signaller}
+}
+
+func (lb *nginxLoadBalancer) Start() error {
+	// Write out initial configuration
+	lb.update(LoadBalancerUpdate{Entries: []LoadBalancerEntry{}})
```
jsravn: Will this break the existing config of the pod gets restarted? E.g. the nginx will become unable to server anything at startup.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+
+	configLocation := lb.confDir + "/nginx.conf"
+	log.Infof("Using config file %s", configLocation)
+	cmd := exec.Command(lb.binary, "-c", configLocation)
+
+	cmd.Stdout = os.Stdout
+	cmd.Stdin = os.Stdin
```
jsravn: I think we should redirect this into our logrus logger, so we get timestamps and the same log format.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+
+func (lb *nginxLoadBalancer) Stop() error {
+	log.Info("Shutting down process %s", lb.cmd.Process.Pid)
+	err := lb.signaller.Sigquit(lb.cmd.Process)
+	if err != nil {
+		log.Error("Error shutting down nginx", err)
+		return err
+	}
+	return nil
+}
```
jsravn: missing newline after brace

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+	err := ioutil.WriteFile(location, contents, 0644)
+	if err != nil {
+		log.Warn("Unable to write new nginx config", err)
+		return false, err
+	}
+	return true, nil
+}
+
+func (lb *nginxLoadBalancer) createConfig(update LoadBalancerUpdate) ([]byte, error) {
+	tmpl, err := template.New("nginx.tmpl").ParseFiles("./nginx.tmpl")
```
jsravn: Shouldn't it be using confDir as a relative path for the template file?

ingress/loadbalancer_util.go
```diff
@@ -0,0 +1,39 @@
...
+		return false
+	}
+	if entry.ServicePort == 0 {
+		return false
+	}
+	return true
+}
+
+// FilterInvalidEntries returns a slice of all the valid LoadBalancer entries
+func FilterInvalidEntries(entries []LoadBalancerEntry) []LoadBalancerEntry {
```
jsravn: Should be lowercase - private.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
 }
 
-func (lb *noopLoadBalancer) Update(entries []LoadBalancerEntry) error {
-	log.Debugf("Updating loadbalancer for %v", entries)
-	return nil
+// Used for generating nginx cofnig
+type loadBalancerTemplate struct {
+	Port       int
+	Entries    []LoadBalancerEntry
+	WorkingDir string
```
jsravn: This gets assigned the confDir, so it's not the working dir. nginx sticks the pidfile in there - not sure it makes sense to do that. Does it even need a pidfile?

cmd/feed-ingress/main.go
```diff
@@ -14,22 +14,28 @@ import (
...
+		defaultTokenFile    = "/run/secrets/kubernetes.io/serviceaccount/token"
+		defaultNginxConfDir = "."
+		defaultNginxPort    = 80
 	)
 
 	flag.StringVar(&apiServer, "apiserver", defaultAPIServer, "Kubernetes API server URL")
 	flag.StringVar(&caCertFile, "cacertfile", defaultCaCertFile, "file containing kubernetes ca certificate")
 	flag.StringVar(&tokenFile, "tokenfile", defaultTokenFile, "file containing kubernetes client authentication token")
+	flag.StringVar(&nginxConfDir, "nginx-conf-dir", defaultNginxConfDir, "directory to store nginx conf")
+	flag.IntVar(&nginxPort, "nginx-port", defaultNginxPort, "nginx port")
```
jsravn: Think it needs better description, such as `port to serve ingress traffic on`. It's not nginx specific either, in case we have different lbs, so I'd rename the flag to `ingress-port` or something like that.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+		return []byte{}, err
+	}
+
+	return output.Bytes(), nil
+}
+
+func diff(b1, b2 []byte) (data []byte, err error) {
+	f1, err := ioutil.TempFile("", "")
+	if err != nil {
+		return
```
jsravn: Isn't it clearer to explicitly return rather than use named return vars? At least I haven't seen that being done much in any golang code.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+	defer f2.Close()
+
+	f1.Write(b1)
+	f2.Write(b2)
+
+	data, err = exec.Command("diff", "-u", f1.Name(), f2.Name()).CombinedOutput()
+	if len(data) > 0 {
+		err = nil
+	}
+	return
```
jsravn: It's hard to tell what gets returned here. If it was explicit, e.g. return data, nil it's obvious there is no error if we get to this point in the function.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+		log.Info("Unable to read existing nginx config file. Assuming it needs creating for the first time.", err)
+		return writeFile(lb.confDir+"/nginx.conf", file)
+
+	}
+	diffOutput, err := diff(existing, file)
+	if err != nil {
+		log.Warn("Unable to diff nginx files", err)
+		return false, err
+	}
+	log.Infof("Diff output %s", string(diffOutput))
```
jsravn: Is there a point to logging the diff if it's empty? Can move this until after the next conditional.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+	if len(diffOutput) == 0 {
+		log.Info("Configuration has not changed")
+		return false, nil
+	}
+
+	log.Info("Configuration is different. Updating configuration.")
+	_, err = writeFile(lb.confDir+"/nginx.conf", file)
+	if err != nil {
+		log.Warn("Unable to write new nginx configuration", err)
```
jsravn: This seems pretty severe and should be an Error I think. Means our nginx is likely broken.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+		return false, err
+	}
+	return true, nil
+}
+
+func writeFile(location string, contents []byte) (bool, error) {
+	err := ioutil.WriteFile(location, contents, 0644)
+	if err != nil {
+		log.Warn("Unable to write new nginx config", err)
```
jsravn: Redundant log.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+		return false, err
+	}
+	return true, nil
+}
+
+func (lb *nginxLoadBalancer) createConfig(update LoadBalancerUpdate) ([]byte, error) {
+	tmpl, err := template.New("nginx.tmpl").ParseFiles("./nginx.tmpl")
+	if err != nil {
+		log.Warn("Unable to generate nginx configurtion. It will be out of date.", err)
```
jsravn: I feel like logging of errors should happen at top level - with whatever is calling this. Instead of logging, expand the error with `fmt.Errorf`

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+		return []byte{}, err
```
jsravn: return `nil, err` instead of an empty byte array

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+	}
+
+	var output bytes.Buffer
+	validEntries := FilterInvalidEntries(update.Entries)
+	err = tmpl.Execute(&output, loadBalancerTemplate{Port: lb.port, Entries: validEntries, WorkingDir: lb.confDir})
+
+	if err != nil {
+		log.Warn("Unable to execute nginx config duration. It will be out of date.", err)
+		return []byte{}, err
```
jsravn: `nil, err`

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+		log.Warn("Unable to generate nginx configurtion. It will be out of date.", err)
+		return []byte{}, err
+	}
+
+	var output bytes.Buffer
+	validEntries := FilterInvalidEntries(update.Entries)
+	err = tmpl.Execute(&output, loadBalancerTemplate{Port: lb.port, Entries: validEntries, WorkingDir: lb.confDir})
+
+	if err != nil {
+		log.Warn("Unable to execute nginx config duration. It will be out of date.", err)
```
jsravn: Same as above, expand the error and let top level take care of logging a problem.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+}
+func (lb *nginxLoadBalancer) Update(entries LoadBalancerUpdate) (bool, error) {
+	updated, err := lb.update(entries)
+	if err != nil {
+		return updated, err
+	}
+	if updated {
+		err = lb.signaller.Sighup(lb.cmd.Process)
+		if err != nil {
+			log.Warn("Configuration was wirtten but unable to signal nginx to reload configuration", err)
```
jsravn: Probably should be `Error`.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+			return false, err
```
jsravn: expand the error instead of logging here.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+	if err != nil {
+		log.Error("Error shutting down nginx", err)
+		return err
+	}
+	return nil
+}
+func (lb *nginxLoadBalancer) Update(entries LoadBalancerUpdate) (bool, error) {
+	updated, err := lb.update(entries)
+	if err != nil {
+		return updated, err
```
jsravn: Isn't it clearer to do `return false, err` ?

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+	if err != nil {
+		log.Error("Unable to start nginx", err, cmd)
+		return err
+	}
+
+	go func() {
+		log.Info("Nginx has exited: ", cmd.Wait())
+	}()
+
+	log.Infof("Nginx pid %s", cmd.Process.Pid)
```
jsravn: Probably be debug. Also, use `%d` for ints not `%s`.

ingress/loadbalancer.go
```diff
@@ -17,19 +37,181 @@ type LoadBalancerEntry struct {
...
+	return nil
+}
+
+func (lb *nginxLoadBalancer) Stop() error {
+	log.Info("Shutting down process %s", lb.cmd.Process.Pid)
```
jsravn: Use `%d`

ingress/loadbalancer.go
jsravn: https://github.com/golang/go/wiki/CodeReviewComments#naked-returns agrees with me :).

jsravn:
Ok done, whew :).

In summary, all looks functionality good in the golang side.
- I think the `nginx.conf` needs some work. I think we should try to start with something prod-like that we can tweak later. I also think we need more configurability so we don't need to change it too much when creating the pod.yaml (like the worker processes). I put my questions/thoughts on the nginx.conf comments.
- Most of the rest is just stylistic things. Some redundant logging, organization of the go files, stuff like that. Tests all look good to me.
- Should run as `nginx` user in Dockerfile - as a basic level of security.

.travis.yml
```diff
@@ -6,3 +6,18 @@ go:
...
+services:
+  - docker
+
+after_success:
+  - docker login -e $DOCKER_EMAIL -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
+  - export REPO=skycirrus/feed-ingress
+  - export TAG=`if [ "$TRAVIS_BRANCH" == "master" ]; then echo "latest"; else echo $TRAVIS_BRANCH ; fi`
+  - cd docker && docker build -f Dockerfile -t $REPO:$COMMIT .
+  - docker tag $REPO:$COMMIT $REPO:$TAG
+  - docker push $REPO
```
jsravn: Can we update `README.md` with details on how the release process works?

jsravn: Also would it make sense to stick this in a shell script so we can run it locally?

.gitignore
```diff
@@ -1,2 +1,3 @@
 .idea
 *.iml
+*.swp
```
jsravn: Silly vim user

cmd/feed-ingress/main.go
chbatey: Done

Makefile
chbatey: Yep. The docker task in the make file now only copies the relevant files to the docker dir. It doesn't build the image.

docker/Dockerfile
chbatey: I don't think it'll break anything else, this sym links those files to stdout which'll mean the nginx logs to to docker.

docker/Dockerfile
chbatey: Oops. vi mode and intellij seems to cause this :-(

ingress/controller.go
chbatey: Agreed

ingress/loadbalancer_util.go
chbatey: Ya, that makes sense.

ingress/loadbalancer_util.go
chbatey: I like that. Avoids us having a huge loadbalancer file :)

docker/Dockerfile
jsravn: Ah okay was reading it backwards. Would it be nicer if we just specify /dev/stdout in the nginx.conf?

jsravn:
I don't mind leaving the nginx.conf tuning alone, we can do it as a separate PR.

jsravn:
@balooo do you want to have a look as well?

ingress/nginx.tmpl
chbatey: Done

ingress/nginx.tmpl
chbatey: ya

ingress/nginx.tmpl
chbatey: I am going to create a task for nginx config tuning

ingress/nginx.tmpl
chbatey: ya that is better

ingress/nginx.tmpl
chbatey: Gateway error happens out of the box. I can add a 404

chbatey: added

ingress/nginx.tmpl
chbatey: i will leave this to you when doing the nginx health check

ingress/nginx.tmpl
chbatey: I thin we should do this when we integrate with an ELB / show it is needed. As soon as we have the ELB controller / verification app setup we can do this

ingress/nginx.tmpl
chbatey: will address as part of the nginx config story

ingress/loadbalancer.go
chbatey: ya, good idea to keep out files smaller. Done.

ingress/loadbalancer.go
chbatey: ya it is created from main.go

ingress/loadbalancer.go
chbatey: yes it was very flakey

cmd/feed-ingress/main.go
chbatey: I think we should just add it to the controller stop

ingress/loadbalancer.go
chbatey: Hrmmmmmmmmmmmmmm

Indeed. Perhaps we should just include a default configuration in the container that is picked on initial start. I will create a separate task for this and work on it after this is PRed

ingress/loadbalancer.go
chbatey: looks like we can use logrus as a Writer so will do this

ingress/loadbalancer.go
chbatey: It could be, as the make copies these into the same dir and the docker image puts them in the same dir this hasn't been required so far

ingress/loadbalancer_util.go
chbatey: done

ingress/loadbalancer.go
chbatey: yeah the conf dir is used for the pid as well, I finally remember why I called it working dir :)

cmd/feed-ingress/main.go
chbatey: reanmed to ingress-port

ingress/loadbalancer.go
chbatey: changed

ingress/loadbalancer.go
chbatey: done

ingress/loadbalancer.go
chbatey: agreed

ingress/loadbalancer.go
chbatey: Removed

ingress/loadbalancer.go
chbatey: I think that is a good practise

ingress/loadbalancer.go
chbatey: doh, done

ingress/loadbalancer.go
chbatey: doh, done

ingress/loadbalancer.go
chbatey: done

ingress/loadbalancer.go
chbatey: done

.travis.yml
chbatey: i haven't found a need for the docker image locally yet, will add release information to the readme

chbatey:
@jsravn i think i have addressed all your comments. I will squash commits.

PR #7, opened 2016-05-17
Add watching of ingresses

jsravn:
The logic tries its best to stay connected while watching. It does this by retrying the connection to the apiserver if it gets disconnected.
    
The watcher also now sends an update on any successful connect, to ensure the controller will not miss updates from reconnects.

### DIFF
--- a/k8s/client.go
+++ b/k8s/client.go
@@ -16,12 +16,34 @@ import (
 	"net/url"
 	"strings"
 
+	"bufio"
+
+	"time"
+
+	"net"
+
+	"math/rand"
+
+	"io"
+
 	log "github.com/Sirupsen/logrus"
+)
+
+const (
+	ingressPath       = "/apis/extensions/v1beta1/ingresses"
+	initialRetryDelay = time.Millisecond * 100
+	maxRetryDelay     = time.Second * 60
 )
 
 // Client for connecting to a Kubernetes cluster.
 type Client interface {
+	// GetIngresses returns all the ingresses in the cluster.
 	GetIngresses() ([]Ingress, error)
+	// WatchIngresses watches for updates to ingresses and notifies the Watcher.
+	// It will also send a notification whenever it connects to the API server,
+	// to notify that there may be existing ingresses that need to be retrieved.
+	// It's intended that client code will call GetIngresses to retrieve
+	// the current state when notified.
 	WatchIngresses(Watcher) error
 }
 
@@ -45,8 +67,15 @@ func New(apiServerURL string, caCert []byte, token string) (Client, error) {
 		return nil, fmt.Errorf("unable to parse ca certificate")
 	}
 
+	// same as net.DefaultTransport, with k8s CAs added
 	tr := &http.Transport{
 		TLSClientConfig: &tls.Config{RootCAs: pool},
+		Dial: (&net.Dialer{
+			Timeout:   10 * time.Second,
+			KeepAlive: 30 * time.Second,
+		}).Dial,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
 	}
 	httpClient := &http.Client{Transport: tr}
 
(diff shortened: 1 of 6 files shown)


jsravn:
@chbatey please have a look

k8s/client.go
```diff
@@ -62,46 +89,198 @@ func New(apiServerURL string, caCert []byte, token string) (Client, error) {
 }
 
 func (c *client) GetIngresses() ([]Ingress, error) {
-	endpoint := c.baseURL + "/apis/extensions/v1beta1/ingresses"
-	req, err := http.NewRequest("GET", endpoint, nil)
+	var ingressList IngressList
```
chbatey: Any reason for this being an inout parameter rather than a return parameter?

k8s/client.go
```diff
@@ -62,46 +89,198 @@ func New(apiServerURL string, caCert []byte, token string) (Client, error) {
...
+		var reconnect bool
+
+		select {
+		case <-w.Done():
+			log.Debug("Watcher is done, stopping watch")
+			return false
+		case update := <-updateCh:
+			if update == nil {
+				log.Info("Long poll terminated, will reconnect")
+				reconnect = true
```
chbatey: Does the break come out of the fall loop? If so could this return there instead of breaking?

k8s/client.go
```diff
@@ -62,46 +89,198 @@ func New(apiServerURL string, caCert []byte, token string) (Client, error) {
...
+	}()
+
+	resp := <-respCh
+	if resp == nil {
+		return nil, fmt.Errorf("request shutdown")
 	}
+	return resp, nil
+}
+
+func (c *client) unmarshallRequest(path string, val interface{}) error {
```
chbatey: This also makes the request where as from the name I expected it to just take a Http response and unmarshall it. Would it make sense to inline request into this method and rename this one to request?

k8s/client.go
```diff
@@ -62,46 +89,198 @@ func New(apiServerURL string, caCert []byte, token string) (Client, error) {
...
+	return ingressList.Items, nil
+}
 
-	log.Debugf("k8s<-: %v", *req)
+func (c *client) WatchIngresses(w Watcher) error {
+	log.Debug("Adding watcher for ingresses")
 
-	resp, err := c.http.Do(req)
+	go func() {
+		ingressRequest := func() (*http.Response, error) {
```
chbatey: Would it be simpler to have this replace client.request?

chbatey: Ah I see request is used without this wrapper else where

k8s/client.go
```diff
@@ -62,46 +89,198 @@ func New(apiServerURL string, caCert []byte, token string) (Client, error) {
...
+	log.Infof("Watching %v", resp.Request.URL)
 
-	if resp.StatusCode < 200 || 300 <= resp.StatusCode {
-		return nil, fmt.Errorf("GET %s returned %v", endpoint, resp)
+	// send an update for a successful watch start
+	w.Updates() <- struct{}{}
+
+	updateCh := make(chan interface{})
+
+	go func() {
```
chbatey: I can we give this a name like handleLongPoll?

chbatey:
This could just be my unfamiliarity with golang but i'd really like to see some identifier in the log messages for what go routine they are on. This isn't something that go supports but what I really want to be able to see from the logs is what each conceptual thread of execution is doing.

From the code review I can see we have:

Execution 1: Controller kicking off the watch and then reading from the Watch channel 

Execution 2: Client kicking off the watch and then waiting for an update on updateCh

Execution 3: Waiting on the long poll and then sending a message to Execution 2.

So not really a per go routine log but our three main threads of execution.

Would it be over kill / error prone to have a separate logger for each of these?

k8s/client.go
```diff
@@ -45,8 +65,15 @@ func New(apiServerURL string, caCert []byte, token string) (Client, error) {
 		return nil, fmt.Errorf("unable to parse ca certificate")
 	}
 
+	// same as net.DefaultTransport, with k8s CAs added
 	tr := &http.Transport{
 		TLSClientConfig: &tls.Config{RootCAs: pool},
+		Dial: (&net.Dialer{
+			Timeout:   30 * time.Second,
```
chbatey: This is a large timeout. Is this because we long poll and thus receive no data on the socket?

k8s/client.go
```diff
@@ -62,46 +89,198 @@ func New(apiServerURL string, caCert []byte, token string) (Client, error) {
...
+	}()
+
+	return nil
+}
+
+type genericList struct {
+	ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+}
+
+func (c *client) getIngressResourceVersion() (string, error) {
```
chbatey: Same as above. We could just return the call to unmarshallRequest

k8s/client.go
jsravn: Will change timeout to 10s.

k8s/client.go
jsravn: Will do.

k8s/client.go
jsravn: So we can have a generic json unmarshaller function.

k8s/client.go
jsravn: Yep will do.

k8s/client.go
jsravn: Ok will rename / split it up.

PR #8, opened 2016-05-20
Add health port to ingress controller

jsravn:
Adds /health and /debug/pprof.

Also:
- Properly os.Exit(-1) if the controller has an error
- Controller controls load balancer lifecycle start
- Make controller start/stop safe, and validate controller state

### DIFF
--- a/cmd/feed-ingress/main.go
+++ b/cmd/feed-ingress/main.go
@@ -6,6 +6,12 @@ import (
 	"os/signal"
 
 	"io/ioutil"
+	"net/http"
+	_ "net/http/pprof"
+
+	"strconv"
+
+	"io"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/sky-uk/feed/ingress"
@@ -20,6 +26,7 @@ var (
 	caCertFile           string
 	tokenFile            string
 	debug                bool
+	healthPort           int
 )
 
 func init() {
@@ -30,6 +37,7 @@ func init() {
 		defaultNginxConfDir = "."
 		defaultIngressPort  = 80
 		defaultNginxWorkers = 1
+		defaultHealthPort   = 12001
 	)
 
 	flag.StringVar(&apiServer, "apiserver", defaultAPIServer, "Kubernetes API server URL")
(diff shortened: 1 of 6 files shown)


cmd/feed-ingress/main.go
```diff
@@ -62,6 +71,19 @@ func configureLogging() {
 	}
 }
 
+func configureHealthPort() {
+	http.HandleFunc("/health", checkHealth)
```
chbatey: Should we shut this down gracefully?

chbatey:
LGTM apart from one question

cmd/feed-ingress/main.go
jsravn: I'll check.

Also, do you think we should wait for nginx to start up before returning healthy?

Should we distinguish between ready and liveness?

jsravn: It doesn't look possible in golang, see https://github.com/golang/go/issues/4674.

We can do it using a third party library like https://github.com/facebookgo/httpdown, but it will add a lot more code. Is it necessary for the health check?

chbatey: No i don't think so

chbatey: Regarding nginx. Yes I think we should wait until we've called Start() on the LB at least. But not if it makes things too complicated

ingress/controller.go
```diff
@@ -104,3 +113,24 @@ func (c *controller) Stop() {
...
 	close(c.stop)
 }
+
+func (c *controller) Healthy() bool {
+	return c.started.get()
+}
+
+type safeBool struct {
+	lock sync.Mutex
+	val  bool
```
chbatey: lol so go doesn't have a concept of a volatile variable :-/

I started reading this and it told me to stop trying to be clever and use a mutex :)

chbatey:
LGTM

jsravn:
I realized we need to handle if the health port / http server dies. I then realized we need to shutdown the controller gracefully. Which made me then realize we don't check error codes on the controller, and return 0 exit code even if the controller had a problem. Whoops! I'm going to fix these issues as part of this PR.

cmd/feed-ingress/main.go
```diff
@@ -102,3 +106,42 @@ func readFile(path string) []byte {
 	}
 	return data
 }
+
+func configureHealthPort(controller ingress.Controller) {
+	http.HandleFunc("/health", checkHealth(controller))
+
+	go func() {
+		log.Error(http.ListenAndServe("localhost:"+strconv.Itoa(healthPort), nil))
```
chbatey: Will this be hit by kubelet? If so localhost won't work as it'll be inside a container

chbatey:
lgtm apart from one question regarding listen address

cmd/feed-ingress/main.go
jsravn: Good point, I'll check it.

jsravn: Ok I changed it to listen on all interfaces. Also changed the port since 12001 is already used in `/etc/services` and it was showing up as some IBM service port.

ingress/controller.go
jsravn: I can probably get rid of this now that I'm locking in start/stop.

PR #9, opened 2016-05-24
Integrate nginx with health checks

jsravn:
- Fail nginx if it dies shortly after startup
- Place all nginx files in working directory
- Check health of loadbalancer on /health endpoint


cmd/feed-ingress/main.go
```diff
@@ -31,23 +31,23 @@ var (
...
-		defaultCaCertFile   = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
-		defaultTokenFile    = "/run/secrets/kubernetes.io/serviceaccount/token"
-		defaultNginxConfDir = "."
-		defaultIngressPort  = 80
-		defaultNginxWorkers = 1
-		defaultHealthPort   = 12082
+		defaultAPIServer       = "https://kubernetes:443"
+		defaultCaCertFile      = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+		defaultTokenFile       = "/run/secrets/kubernetes.io/serviceaccount/token"
+		defaultNginxWorkingDir = "/"
```
chbatey: Did . not make more sense?

ingress/nginx.go
```diff
@@ -33,43 +40,63 @@ type loadBalancerTemplate struct {
...
+	err := os.Remove(lb.nginxConfFile())
+	if err != nil {
+		log.Debugf("Can't remove nginx.conf: %v", err)
+	}
+	_, err = lb.update(LoadBalancerUpdate{Entries: []LoadBalancerEntry{}})
+	return err
+}
+
 func (lb *nginxLoadBalancer) Stop() error {
-	log.Infof("Shutting down process %d", lb.cmd.Process.Pid)
```
chbatey: Was this causing a race? This was useful to know when I ended up with a zombie :)

chbatey:
lgtm with a couple of questions

cmd/feed-ingress/main.go
jsravn: I was thinking defaults should match whatever it runs in the docker container as. So I thought it would be '/' in container. But actually I've put it in '/nginx' because it creates a ton of temp files. It doesn't really matter what the default is. I could swap it to '/nginx'. Were you using '.'?

ingress/nginx.go
jsravn: Just seemed a bit pointless to print out the pid there. Is it useful?

PR #10, opened 2016-05-25
Use nginx user and move release into Makefile

jsravn:
Also change it to only release on master builds. To limit number of
images we have (I don't think we need branch builds? We can always push
it locally if we really need it).

And change default ingress port to 8080 so we don't need super user rights
to start up nginx. Needed so we can run as nginx in the Dockerfile.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,13 +11,4 @@ services:
   - docker
 
 after_success:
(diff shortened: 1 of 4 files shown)


chbatey:
lgtm after travis is happy

PR #11, opened 2016-05-25
Graceful termination of nginx

jsravn:
Two separate commits - please don't squash them.

First one fixes loadbalancer to terminate nginx gracefully.

Second fixes controller to shutdown gracefully.

### DIFF
--- a/cmd/feed-ingress/main.go
+++ b/cmd/feed-ingress/main.go
@@ -13,6 +13,8 @@ import (
 
 	"io"
 
+	"syscall"
+
 	log "github.com/Sirupsen/logrus"
 	"github.com/sky-uk/feed/ingress"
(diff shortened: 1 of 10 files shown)


cmd/feed-ingress/main.go
```diff
@@ -67,7 +69,7 @@ func main() {
 		os.Exit(-1)
 	}
 
-	os.Exit(0)
+	select {}
```
chbatey: What does this do?

chbatey:
lgtm

cmd/feed-ingress/main.go
jsravn: Blocks forever, since I made `controller.Start()` non blocking now (so we can handle shutdown in the signal handler).

PR #13, opened 2016-05-26
Check formatting in travis

jsravn:
Fail the build if files are not formatted.

Also organized the Makefile to make it a bit clearer the various
targets.


chbatey:
Good idea. LGTM

PR #16, opened 2016-05-26
Only release on a master build

jsravn:
Before it was including PRs that were being tested against master.


chbatey:
This makes sense that we don't do the release for PRs :) Good catch

PR #17, opened 2016-05-27
Fix nginx permissions

jsravn:
Fixes nginx permissions so the container can start up.

Also moved all ephemeral files into `build` so it's easy to remove.


balooo:
lgtm

PR #18, opened 2016-05-27
Print nginx version on startup

jsravn:
So we know what version is running in our clusters (for security
updates, and things).


balooo:
lgtm

PR #19, opened 2016-05-31
Reorganize ingress library

jsravn:
Put nginx in its own package, move SafeBool into a util package. Put
common types into ingress/api.


chbatey:
lgtm

PR #23, opened 2016-05-31
initial feed-dns

balooo:
Work related to https://github.com/sky-uk/umc-core/issues/772

### DIFF
--- a/cmd/feed-dns/main.go
+++ b/cmd/feed-dns/main.go
@@ -7,7 +7,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/sky-uk/feed/dns"
 	"github.com/sky-uk/feed/k8s"
-	"github.com/sky-uk/feed/util"
+	"github.com/sky-uk/feed/util/cmd"
 )
 
 var (
@@ -35,13 +35,13 @@ func init() {
 
 func main() {
 	flag.Parse()
-	util.ConfigureLogging(debug)
+	cmd.ConfigureLogging(debug)
 
 	client := k8s.New(caCertFile, tokenFile, apiServer)
 	controller := dns.New(client)
 
-	util.ConfigureHealthPort(controller, healthPort)
-	util.AddSignalHandler(controller)
+	cmd.ConfigureHealthPort(controller, healthPort)
+	cmd.AddSignalHandler(controller)
 
 	err := controller.Start()
 	if err != nil {
--- a/cmd/feed-dns/main.go
+++ b/cmd/feed-dns/main.go
@@ -3,21 +3,11 @@ package main
 import (
 	"flag"
 	"os"
-	"os/signal"
-
-	"io/ioutil"
-	"net/http"
-	_ "net/http/pprof"
-
-	"strconv"
-
-	"io"
-
-	"syscall"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/sky-uk/feed/dns"
 	"github.com/sky-uk/feed/k8s"
+	"github.com/sky-uk/feed/util"
 )
 
 var (
@@ -45,12 +35,13 @@ func init() {
 
 func main() {
 	flag.Parse()
-	configureLogging()
+	util.ConfigureLogging(debug)
 
-	controller := dns.New(createK8sClient())
+	client := k8s.New(caCertFile, tokenFile, apiServer)
+	controller := dns.New(client)
 
-	configureHealthPort(controller)
-	addSignalHandler(controller)
+	util.ConfigureHealthPort(controller, healthPort)
+	util.AddSignalHandler(controller)
 
 	err := controller.Start()
 	if err != nil {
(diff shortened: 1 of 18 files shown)


cmd/feed-dns/main.go
```diff
@@ -0,0 +1,133 @@
...
+		if !controller.Healthy() {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			io.WriteString(w, "fail\n")
+			return
+		}
+
+		w.WriteHeader(http.StatusOK)
+		io.WriteString(w, "ok\n")
+	}
+}
```
jsravn: We should move this handler into `util` and share it.

cmd/feed-dns/main.go
```diff
@@ -0,0 +1,133 @@
...
+			log.Infof("Signalled %v, shutting down gracefully", sig)
+			err := controller.Stop()
+			if err != nil {
+				log.Error("Error while stopping controller: ", err)
+				os.Exit(-1)
+			}
+			os.Exit(0)
+		}
+	}()
+}
```
jsravn: Same here, we should avoid duplication with `feed-ingress`.

k8s/client.go
```diff
@@ -53,7 +55,29 @@ type client struct {
...
 // New creates a client for the kubernetes apiserver.
-func New(apiServerURL string, caCert []byte, token string) (Client, error) {
+func New(caCertFile string, tokenFile string, apiServer string) Client {
+	caCert := readFile(caCertFile)
+	token := string(readFile(tokenFile))
+
+	client, err := createK8sClient(apiServer, caCert, token)
+	if err != nil {
+		log.Errorf("Unable to create Kubernetes client: %v", err)
+		os.Exit(-1)
```
jsravn: Libraries should not be calling `os.Exit`. Only main.go should ever exit to the OS. Instead, return an error and let the main code handle it as it wants.

k8s/client.go
```diff
@@ -53,7 +55,29 @@ type client struct {
 }
 
 // New creates a client for the kubernetes apiserver.
-func New(apiServerURL string, caCert []byte, token string) (Client, error) {
+func New(caCertFile string, tokenFile string, apiServer string) Client {
+	caCert := readFile(caCertFile)
+	token := string(readFile(tokenFile))
```
jsravn: I don't agree moving file reading inside of the k8s library. It makes it less generic and harder to test. The files are more about how the binary itself interacts with an external system. We could just as easily change it to take it as a  string of the token rather than a filename, and I don't think that should require changing the k8s library. Also consider if another project depends on the k8s library, we shouldn't force them to use a file.

util/controller.go
```diff
@@ -0,0 +1,74 @@
...
+	// Run the controller, returning immediately after it starts or an error occurs.
+	Start() error
+	// Stop the controller, blocking until it stops or an error occurs.
+	Stop() error
+	// Healthy returns true for a healthy controller, false for unhealthy.
+	Healthy() bool
+}
+
+// ConfigureHealthPort is used to expose the controllers health over http
+func ConfigureHealthPort(controller Controller, healthPort int) {
```
jsravn: Change this to be a method on controller. And rename to `StartHealthPort`?

util/controller.go
```diff
@@ -0,0 +1,74 @@
...
+	http.HandleFunc("/health", checkHealth(controller))
+
+	go func() {
+		log.Error(http.ListenAndServe(":"+strconv.Itoa(healthPort), nil))
+		log.Info(controller.Stop())
+		os.Exit(-1)
+	}()
+}
+
+func checkHealth(controller Controller) func(w http.ResponseWriter, r *http.Request) {
```
jsravn: Method on the controller

util/controller.go
```diff
@@ -0,0 +1,74 @@
...
+			return
+		}
+
+		w.WriteHeader(http.StatusOK)
+		io.WriteString(w, "ok\n")
+	}
+}
+
+// AddSignalHandler allows the  controller to shutdown gracefully by respecting SIGTERM
+func AddSignalHandler(controller Controller) {
```
jsravn: Method on the controller, and rename to `RegisterSignalHandler`?

util/controller.go
```diff
@@ -0,0 +1,74 @@
+package util
```
jsravn: Maybe we should rename package to 'base' or 'common' since it's less of a util package now.

chbatey:
Will take a look in the morning

dns/controller.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+type controller struct {
+	client        k8s.Client
+	watcher       k8s.Watcher
+	started       bool
+	startStopLock sync.Mutex
+}
+
+// New creates an dns controller.
```
chbatey: a DNS

dns/controller.go
```diff
@@ -0,0 +1,91 @@
...
+func (c *controller) watchForUpdates() {
+	for {
+		select {
+		case <-c.watcher.Done():
+			return
+		case <-c.watcher.Updates():
+			log.Info("Received update on watcher")
+			err := c.updateDNSRecords()
+			if err != nil {
+				log.Errorf("Unable to update dns records: %v", err)
```
chbatey: Can you put in the error message what the result of this will be e.g any new ingresses won't be resolvable and list them

dns/controller_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"fmt"
+
+	"github.com/sky-uk/feed/k8s"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+const smallWaitTime = time.Millisecond * 50
+
+type fakeClient struct {
```
chbatey: Can we share the mock for the k8s client? It will be a pain to update both everytime we change it

chbatey:
Looks good, just a few comments + the rebase

dns/controller_test.go
jsravn: I don't think you can easily share test code (by intention). We'd have to create non test code to share and depend on testify in it.

dns/controller.go
jsravn: Also error messages should be all lowercase.  https://github.com/golang/go/wiki/CodeReviewComments#error-strings

balooo: I haven't done any processing of the ingress resources yet.

ingress/controller.go
```diff
@@ -10,30 +10,21 @@ import (
 	"fmt"
 
 	log "github.com/Sirupsen/logrus"
-	"github.com/sky-uk/feed/ingress/api"
+	"github.com/sky-uk/feed/api"
+	iApi "github.com/sky-uk/feed/ingress/api"
```
jsravn: Instead of `iApi`, can we rename `ingress/api` to `ingress/types`. Since it's internal to the ingress stuff, it makes sense to call it types here.

jsravn:
Added comment re. `iApi`.

PR #25, opened 2016-05-31
Add ability to specify dns resolver

jsravn:
This is necessary in Kubernetes so we can specify skydns, since the pod
will be running on host networking.


jsravn:
I forgot about search domains. Doh. We could pass in the cluster domain name. WDYT @chbatey

PR #26, opened 2016-05-31
Unhealthy if controller is not watching

jsravn:
Also change the `Healthy() bool` methods to `Health() error` so we can more easily debug issues in running controllers.


balooo:
lgtm

PR #31, opened 2016-06-01
Restrict ingress access

jsravn:
Deny everything but localhost by default.

Add support for `sky.uk/allow` annotation on ingress resources.

Add support for a default allow range on ingress, to support blanket
internal access.

### DIFF
--- a/ingress/nginx/nginx_test.go
+++ b/ingress/nginx/nginx_test.go
@@ -15,7 +15,8 @@ import (
 )
 
 const (
-	port = 9090
+	port         = 9090
+	defaultAllow = "10.50.0.0/16"
 )
 
 type mockSignaller struct {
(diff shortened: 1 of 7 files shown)


ingress/nginx/nginx_test.go
```diff
@@ -108,30 +110,86 @@ func TestReloadOfConfig(t *testing.T) {
 
 	assert.NoError(lb.Start())
 
-	entries := []api.LoadBalancerEntry{
-		api.LoadBalancerEntry{
-			Host:        "chris.com",
-			Path:        "/path",
-			ServiceName: "service",
-			ServicePort: 9090,
+	var tests = []struct {
```
chbatey: Nice

jsravn: sarcasm? haha :)

PR #32, opened 2016-06-02
Rename 'nginx-status-port' to 'ingress-status-port'

jsravn:
Since we'll need this port regardless of what the implementation is.
It's used by frontends to determine health of ingress.


balooo:
lgtm,  feel free to merge after mine

PR #33, opened 2016-06-02
Add ingress names to the nginx.conf

jsravn:
So we can easily find the ingress in kubernetes that generated the
entry.

This also changes it to sort by name so our config order should remain
relatively stable (not be reordered whenever there is a new or changed
ingress).


chbatey:
Good idea. LGTM after rebase

PR #34, opened 2016-06-03
Register with ELB front end

chbatey:
This is not wired up yet just submitting so we can review separately.

Vendored a new dependency as a separate commit.


ingress/types/frontend.go
```diff
@@ -0,0 +1,18 @@
...
+	Attach(frontend FrontendInput) (int, error)
+	// Detach should de-register the external load balancer
+	// It should block until detaching is complete
+	Detach(frontend FrontendInput) error
+}
+
+// FrontendInput contains the information for attaching to a Frontend
+type FrontendInput struct {
+	Cluster string
+}
```
jsravn: Is it necessary to have an additional struct if it's just a single parameter? I get the idea of value types and all but I think it's a bit too much verbosity (same with LoadBalancerUpdate, it's a bit painful having to always wrap the entry array - when it's obvious it's an array of entries).

ingress/types/frontend.go
```diff
@@ -0,0 +1,18 @@
...
+	// Returns the number of front ends attached
+	Attach(frontend FrontendInput) (int, error)
+	// Detach should de-register the external load balancer
+	// It should block until detaching is complete
+	Detach(frontend FrontendInput) error
+}
+
+// FrontendInput contains the information for attaching to a Frontend
+type FrontendInput struct {
+	Cluster string
```
jsravn: Comment what this is used for? Looks for tag `sky.uk/cluster` on elbs or something like that.

ingress/elb/elb_test.go
```diff
@@ -0,0 +1,271 @@
...
+	e.(*elb).maxTagQuery = 1
+	instanceID := "cow"
+	loadBalancerName1 := "lb1"
+	loadBalancerName2 := "lb1"
+	mockInstanceMetadata(mockMetadata, instanceID)
+	mockLoadBalancers(mockElb, loadBalancerName1, loadBalancerName2)
+	mockClusterTags(mockElb, lbTags{name: loadBalancerName1, tags: []*aws_elb.Tag{&aws_elb.Tag{Key: aws.String(frontendTag), Value: aws.String(clusterName)}}})
+	mockClusterTags(mockElb, lbTags{name: loadBalancerName2, tags: []*aws_elb.Tag{&aws_elb.Tag{Key: aws.String(frontendTag), Value: aws.String(clusterName)}}})
+	mockRegisterInstances(mockElb, loadBalancerName1, instanceID)
+	mockRegisterInstances(mockElb, loadBalancerName2, instanceID)
```
jsravn: So many mocks! This scares me for future maintenance. But this class is the abstraction on AWS, so I see why we need to do it to fully test `elbs` behaviour. I guess in the future if the logic becomes more complex (such as with dynamic ELB updating) we may want to extract the AWS calls into a thin wrapper that we can mock instead.

ingress/elb/elb.go
```diff
@@ -0,0 +1,136 @@
...
+			for _, tag := range description.Tags {
+				if *tag.Key == elbTag && *tag.Value == e.clusterName {
+					clusterFrontEnds = append(clusterFrontEnds, *description.LoadBalancerName)
+				}
+			}
+		}
+	}
+
+	for _, frontend := range clusterFrontEnds {
+		// TODO deal with error
```
jsravn: Don't forget! :)

ingress/elb/elb.go
```diff
@@ -0,0 +1,136 @@
...
+
+	}
+	return len(clusterFrontEnds), nil
+}
+
+func (e *elb) Detach(frontend types.FrontendInput) error {
+	return nil
+}
+
+func min(x, y int) int {
```
jsravn: This really doesn't exist in golang? Wow.

ingress/elb/elb.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		for _, description := range output.TagDescriptions {
+			for _, tag := range description.Tags {
+				if *tag.Key == elbTag && *tag.Value == e.clusterName {
+					clusterFrontEnds = append(clusterFrontEnds, *description.LoadBalancerName)
+				}
+			}
+		}
+	}
```
jsravn: Stick the load balancer lookup into its own method.

ingress/elb/elb.go
```diff
@@ -0,0 +1,136 @@
...
+func (e *elb) Attach(frontend types.FrontendInput) (int, error) {
+	log.Info("Attaching to loadbalancer with %v", frontend)
+
+	id, err := e.metadata.GetInstanceIdentityDocument()
+	if err != nil {
+		return 0, fmt.Errorf("unable to query ec2 metadata service for InstanceId: %v", err)
+	}
+
+	instance := id.InstanceID
+	log.Info("Ingress controller running on instance: ", instance)
```
jsravn: "Attaching to ELBs from instance %s", instanceId? Since it's about the elb attaching.

ingress/elb/elb.go
```diff
@@ -0,0 +1,136 @@
...
+			Marker: resp.NextMarker,
+		}
+	}
+
+	var clusterFrontEnds []string
+	totalLbs := len(lbNames)
+	for i := 0; i < len(lbNames); i += e.maxTagQuery {
+		to := min(i+e.maxTagQuery, totalLbs)
+		log.Debugf("Querying tags from %d to %d", i, to)
+		// Go slices are inclusive:exclusive
```
jsravn: Probably don't need this comment

ingress/elb/elb.go
```diff
@@ -0,0 +1,136 @@
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
+	aws_elb "github.com/aws/aws-sdk-go/service/elb"
+	"github.com/sky-uk/feed/ingress/types"
+)
+
+const (
+	elbTag = "KubernetesClusterFrontend"
```
jsravn: Prefix with `sky.uk` like Joe has done.

cmd/feed-ingress/main.go
```diff
@@ -84,13 +85,14 @@ func init() {
 func main() {
 	flag.Parse()
 	cmd.ConfigureLogging(debug)
-
 	lb := createLB()
+	frontend := elb.New("eu-west-1", "test")
```
jsravn: Won't this break if we merge it in to master? Can we add a flag to disable / enable frontends? Either with `--expected-frontends=0`  or by having a blank cluster name or something.

jsravn:
lgtm once comments are addressed

ingress/types/frontend.go
chbatey: I have gotten rid of this in my next branch so it'll be done soon :)

ingress/types/frontend.go
chbatey: It is going away

ingress/elb/elb_test.go
chbatey: I agree. It is due to the craziness of the aws API for querying by tag

ingress/elb/elb.go
chbatey: haha

ingress/elb/elb.go
chbatey: Nope, only for floats and it doesn't have overloading

cmd/feed-ingress/main.go
chbatey: It isn't started yet :) Doing that in a separate PR

chbatey:
Thanks for the review, addresses the comments

PR #35, opened 2016-06-03
Add route53 dependency

balooo:


chbatey:
Where is the change to the vendors file?

chbatey:
Also I think this should just be a commit in your PR that uses this dependency

jsravn:
I think it's okay if this is its own PR, makes it easier to rebase the next PR if it's a lot of commits. I'm indifferent...

jsravn:
I guess it's sort of pointless on its own :).

chbatey:
Either or :)

PR #36, opened 2016-06-03
Deregister with front end ELBs on shutdown

chbatey:


jsravn:
Commit is too big to view in github :). Maybe we should put dependencies in a separate PR.

cmd/feed-ingress/main.go
```diff
@@ -84,13 +90,19 @@ func init() {
 		"Keep alive time for persistent client connections to nginx.")
 	flag.StringVar(&nginxLogLevel, "nginx-loglevel", defaultNginxLogLevel,
 		"Log level for nginx. See http://nginx.org/en/docs/ngx_core_module.html#error_log for levels.")
+	flag.StringVar(&clusterName, "cluster-name", defaultClusterName,
+		"Kubernetes cluster name. Used to find front ends associated with this cluster")
```
jsravn: Can we document the tag name it looks for?

ingress/controller.go
```diff
@@ -70,6 +70,14 @@ func (c *controller) Start() error {
 
 	go c.watchForUpdates()
 
+	frontends, err := c.frontend.Attach()
+
+	if err != nil {
+		return fmt.Errorf("unable to attach to front end %v", err)
+	}
+
+	log.Infof("Attached to %d frontend(s)", frontends)
```
jsravn: Would it be better to just log this in the frontend itself? That way we're not returning a value just so it can be logged, Attach can just return the error.

ingress/controller.go
```diff
@@ -141,6 +149,8 @@ func (c *controller) Stop() error {
 
 	log.Info("Stopping controller")
 
+	c.frontend.Detach()
```
jsravn: Check the error.

ingress/elb/elb.go
```diff
@@ -16,21 +16,28 @@ const (
 )
 
 // New  creates a new frontend
```
jsravn: "creates a new ELB frontend"

ingress/elb/elb.go
```diff
@@ -65,18 +75,33 @@ func (e *elb) Attach(frontend types.FrontendInput) (int, error) {
...
 
+	// Save these now so we can always know what we might have done
+	// up until this point we have only read data
+	e.elbs = clusterFrontEnds
+	e.instanceID = instance
+	registered := 0
+
+	if len(clusterFrontEnds) != e.expectedFrontends {
+		log.Warnf("Expected to find %d elb frontends, only found %d", e.expectedFrontends, len(clusterFrontEnds))
+	}
```
jsravn: If this isn't going to fail at startup, we should report it in a Health() status. That way kubernetes will try to restart / reattach it.

ingress/elb/elb.go
```diff
@@ -65,18 +75,33 @@ func (e *elb) Attach(frontend types.FrontendInput) (int, error) {
...
+		_, err = e.awsElb.RegisterInstancesWithLoadBalancer(&aws_elb.RegisterInstancesWithLoadBalancerInput{
 			Instances: []*aws_elb.Instance{
 				&aws_elb.Instance{
 					InstanceId: aws.String(instance),
 				}},
 			LoadBalancerName: aws.String(frontend),
 		})
 
+		if err != nil {
+			return registered, fmt.Errorf("unable to register instance %s with elb %s: %v", instance, frontend, err)
```
jsravn: Shouldn't it keep trying to register the rest of the ELBs before exiting?

ingress/elb/elb.go
```diff
@@ -131,8 +157,22 @@ func (e *elb) findFrontEndElbs() ([]string, error) {
...
+	var result error
+	for _, elb := range e.elbs {
+		log.Infof("Deregistering instance %s with elb %s", e.instanceID, elb)
+		_, err := e.awsElb.DeregisterInstancesFromLoadBalancer(&aws_elb.DeregisterInstancesFromLoadBalancerInput{
+			Instances:        []*aws_elb.Instance{&aws_elb.Instance{InstanceId: aws.String(e.instanceID)}},
+			LoadBalancerName: aws.String(elb),
+		})
+
+		if err != nil {
+			result = fmt.Errorf("unable to deregister instance %s with elb %s: %v", e.instanceID, elb, err)
```
jsravn: We should probably concat the errors rather than losing them. Like store a list of them and return that. It will be misleading if there are multiple errors.

ingress/elb/elb.go
```diff
@@ -65,18 +75,33 @@ func (e *elb) Attach(frontend types.FrontendInput) (int, error) {
 		return 0, err
 	}
 
+	// Save these now so we can always know what we might have done
+	// up until this point we have only read data
+	e.elbs = clusterFrontEnds
```
jsravn: It may be a bit more robust to only store the ones successfully registered, so when it detaches it will avoid errors. But maybe not a big deal as long as we ignore the errors on detach.

cmd/feed-ingress/main.go
```diff
@@ -96,11 +100,14 @@ func init() {
 		"AWS region")
 	flag.IntVar(&expectedFrontends, "expected-frontends", defaultExpectedFrontends,
 		"Expected number of front ends. If 0 the controller will not attempt to attach. A warning is logged if the incorrect number are found.")
+	flag.StringVar(&metricsListenAddress, "metrics-listen-address", defaultMetricsListenAddress,
+		"Listen address for metrics endpoint. Defaults to :8083")
```
jsravn: Can't it just use the health port? We put pprof debug on there as well.

cmd/feed-ingress/main.go
```diff
@@ -136,3 +143,8 @@ func createLB() types.LoadBalancer {
 		DefaultAllow:      ingressAllow,
 	})
 }
+
+func startMetrics() {
+	http.Handle("/metrics", prometheus.Handler())
+	http.ListenAndServe(metricsListenAddress, nil)
```
jsravn: I'm not sure you can have multiple of these, I think they use global state (so it will conflict with our /health /debug one).

ingress/elb/elb.go
chbatey: +1 I'll do health checks in the next PR

ingress/elb/elb.go
chbatey: Yeah that makes sense

chbatey: Actually the controller fails to start if any of its Start() methods fail so not worth doing this unless we change that

ingress/elb/elb.go
chbatey: Hrmm yeah

cmd/feed-ingress/main.go
chbatey: Updated

cmd/feed-ingress/main.go
chbatey: Changed to use the same one

PR #38, opened 2016-06-06
Create a generic controller

jsravn:
That watches ingress resources from Kubernetes. To be used by both the
dns and ingress controllers.

Also modify the ingress controller to use it.


chbatey:
We now have three files called `controller.go`. Do we still need the utils one? Can the functions be moved to the generic one? Or at least the files named less generically for the specific controllers.

controller/ingress_entry.go
```diff
@@ -28,8 +28,8 @@ type LoadBalancerEntry struct {
 }
 
 // FilterInvalidEntries returns a slice of all the valid LoadBalancer entries
```
chbatey: Ingresses rather than LoadBalancer. 

Do we want each Updater to validate rather than the controller or do you think this is applicable for all?

controller/ingress_entry.go
```diff
@@ -60,14 +60,14 @@ func (entry LoadBalancerEntry) ValidateEntry() bool {
...
 	copy(sortedEntries, u.Entries)
 	sort.Sort(byName(sortedEntries))
-	return LoadBalancerUpdate{Entries: sortedEntries}
+	return IngressUpdate{Entries: sortedEntries}
 }
 
-type byName []LoadBalancerEntry
+type byName []IngressEntry
 
 func (a byName) Len() int           { return len(a) }
```
chbatey: Is it not clearer just to have the user call len?

balooo:
`lb_test.go` should be `ingress_updater_test.go`

ingress/proxy.go
```diff
@@ -0,0 +1,16 @@
+package ingress
+
+import "github.com/sky-uk/feed/controller"
+
+// Proxy proxies traffic from external clients to the internal services in Kubernetes.
+type Proxy interface {
```
chbatey: Why do we need another interface for proxy? Is this just not an implementation of the Upater interface?

jsravn:
> We now have three files called controller.go.

Is it a problem? I think in IntelliJ it shows the package as well, like `ingress/controller.go`. It's a bit annoying in intellij since the tab only shows the filename, but smurf naming controllers isn't great either. If we can think of a better name for controller.go then that's fine as well. I think the cmd one could probably change, or moved into controller.go.

chbatey:
A few comments / questions. Assigning back.

controller/ingress_entry.go
jsravn: This is required for sorting the entries. To sort something you have to implement `Len` and `Swap`. That was done in a separate PR when I ordered nginx entries.

controller/ingress_entry.go
jsravn: For now it's applicable for all of them, although I think it may make more sense to put this logic inside the generic controller rather than standalone on the entry, since it is specific to the controller's logic.

ingress/proxy.go
jsravn: Well we actually had `type Proxy controller.Updater`. I think it's nicer to make it explicitly for proxies rather than reuse updater, otherwise it gets pretty confusing having updaters calling updaters.

But the bool return value on Update doesn't make sense for the dns updater, so the interfaces don't actually match currently.

PR #39, opened 2016-06-06
Add k8s client support for service resources

jsravn:


jsravn:
This is ready for review.

controller/controller.go
```diff
@@ -67,29 +67,19 @@ func (c *controller) Start() error {
...
-			return
-		case <-c.watcher.Updates():
-			log.Info("Received update on watcher")
-			err := c.updateLoadBalancer()
-			if err != nil {
-				log.Errorf("Unable to update load balancer: %v", err)
-			}
+	for _ = range c.watcher.Updates() {
+		log.Info("Received update on watcher")
+		err := c.updateLoadBalancer()
```
balooo: can we rename 'updateLoadBalancer'  to just 'update' or 'doUpdate', but it's not load balancer specific anymore

controller/controller.go
```diff
@@ -67,29 +67,19 @@ func (c *controller) Start() error {
...
+		if err != nil {
+			log.Errorf("Unable to update load balancer: %v", err)
```
balooo: I don't think this is load balancer specific

balooo:
lgtm, 2 minor comments

PR #40, opened 2016-06-07
Frontend health check that asserts the correct number of front ends h…

chbatey:
…ave been attached

I've also moved the gauge into the frontend implementation to make it easier to conform to the Upater interface later.

### DIFF
--- a/ingress/elb/elb.go
+++ b/ingress/elb/elb.go
@@ -5,6 +5,8 @@ import (
 
 	"errors"
 
+	"sync"
+
 	log "github.com/Sirupsen/logrus"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
@@ -49,6 +51,7 @@ type elb struct {
 	instanceID          string
 	elbs                []string
 	registeredFrontends int
+	sync.Mutex
 }
 
 // ELB interface to allow mocking of real calls to AWS as well as cutting down the methods from the real
@@ -68,6 +71,8 @@ type EC2Metadata interface {
 }
 
 func (e *elb) Attach() error {
+	e.Lock()
+	defer e.Unlock()
 
 	if e.expectedFrontends == 0 {
 		return nil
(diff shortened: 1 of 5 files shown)


ingress/elb/elb.go
```diff
@@ -181,6 +190,13 @@ func (e *elb) Detach() error {
 	return nil
 }
 
+func (e *elb) Health() error {
+	if e.registeredFrontends != e.expectedFrontends {
```
jsravn: Doesn't this need to be synchronized?

jsravn: Thought about it a bit more - I think it does, since the goroutine that starts the http server for health checks may happen before the attach call. We can control that order in the main but I think it's safer just to assume health checks are thread safe.

ingress/ingress_updater.go
```diff
@@ -29,15 +21,12 @@ func New(frontend Frontend, proxy Proxy) controller.Updater {
 }
 
 func (u *updater) Start() error {
-	prometheus.Register(attachedFrontendGauge)
 
-	frontEnds, err := u.frontend.Attach()
+	_, err := u.frontend.Attach()
```
jsravn: Can we get rid of the int now? It's an implementation detail of the elb frontend now.

ingress/ingress_updater_test.go
```diff
@@ -117,11 +125,27 @@ func TestUpdaterReturnsHealthOfProxy(t *testing.T) {
...
+	frontend.On("Attach").Return(1, nil)
+	frontend.On("Health").Return(nil).Once()
+	frontend.On("Health").Return(errors.New("Oh dear oh dear"))
 
 	assert := assert.New(t)
 	assert.NoError(controller.Start())
 
 	assert.NoError(controller.Health(), "first it's healthy")
 	assert.Error(controller.Health(), "then it's not")
+
```
jsravn: blank line, delete :)

chbatey:
Updated based on your review comments, just waiting for a clean build

ingress/elb/elb.go
chbatey: And to provide visibility to the write. I will add a lock around Attach and Health

PR #41, opened 2016-06-09
Use service IP for nginx proxy backend

jsravn:
Changes:
- Subscribe to service updates as well
- Remove the resolver logic, nginx just uses the IP
- Filtering logic encapsulated in the generic controller, so the dns
  controller gets the same entries as nginx.


jsravn:
This is ready for review.

controller/controller.go
```diff
@@ -26,33 +26,31 @@ type Controller interface {
...
-	updater       Updater
-	client        k8s.Client
-	watcher       k8s.Watcher
-	started       bool
-	startStopLock sync.Mutex
-	serviceDomain string
+	updater     Updater
+	client      k8s.Client
+	watcher     k8s.Watcher
+	watcherDone sync.WaitGroup
```
chbatey: As we only ever add 1 to this (I think) we could use a condition variable. WDYT?

chbatey:
One question otherwise lgtm

PR #42, opened 2016-06-10
Failed to start if expected frontends not matched

chbatey:
Doh, all of the other demo comments were actually in umc-platform so should have just snook this in the last one

### DIFF
--- a/elb/elb.go
+++ b/elb/elb.go
@@ -5,8 +5,6 @@ import (
 
 	"errors"
 
-	"sync"
-
 	log "github.com/Sirupsen/logrus"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
@@ -51,7 +49,6 @@ type elb struct {
 	instanceID          string
 	elbs                []string
 	registeredFrontends int
-	sync.Mutex
 }
 
 // ELB interface to allow mocking of real calls to AWS as well as cutting down the methods from the real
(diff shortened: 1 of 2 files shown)


ingress/elb/elb.go
```diff
@@ -119,6 +119,11 @@ func (e *elb) Attach() error {
 	prometheus.Register(attachedFrontendGauge)
 	attachedFrontendGauge.Set(float64(registered))
 	e.registeredFrontends = registered
+
+	if registered != e.expectedFrontends {
+		log.Warn("Expected ELBs: %d actual: %d", e.expectedFrontends, registered)
```
balooo: should be `log.Warnf`

balooo:
Also, if we're checking that registered == expected on startup,  I don't understand why we are checking it as part of the Health check.   If it fails on startup, it won't startup.  If we want to check those values on a healthcheck poll, shouldn't we re query the number of registered to make sure it still matches?

jsravn:
This made me think we should just fail if it can't attach. Because the deployment will fail if it never goes healthy (it waits for the pod to be ready, otherwise deployment times out). So we may as well fail the startup if it doesn't match rather than just leaving it to the health check.

jsravn:
Similarly for the Health check, we should probably periodically check the frontends are still attached. But that can be a separate task. (And maybe it's too edge case to worry about..?)

chbatey:
yeah let's just return an errro

chbatey:
There's still value in logging this even if kubelet does just kill it as we will get to see why it  killed it from the last container's logs

jsravn:
If it returns an error on `Attach` it will get logged by `main.go`.

chbatey:
I have updated this to return an error from Start() if the number is not matched. Ready for review.

jsravn:
lgtm

PR #45, opened 2016-06-10
Change controller.go to take multiple updaters

jsravn:
This simplifies the logic by removing the need for
`ingress-updater.go`. Instead the generic controller simply operates on
a slice of Updaters.

Also improve `util/cmd` by changing it to not depend on the
controller package.

And improve the controller health check by reporting if updates failed.


controller/controller.go
```diff
@@ -169,8 +177,10 @@ func (c *controller) Stop() error {
 	close(c.watcher.Done())
 	c.watcherDone.Wait()
 
-	if err := c.updater.Stop(); err != nil {
-		log.Warnf("Error while stopping: %v", err)
+	for _, u := range c.updaters {
+		if err := u.Stop(); err != nil {
+			log.Warnf("Error while stopping: %v", err)
```
chbatey: Should we log the updater and make sure each updater implements string()

controller/controller.go
```diff
@@ -186,13 +196,19 @@ func (c *controller) Health() error {
 		return fmt.Errorf("controller has not started")
 	}
 
-	if err := c.updater.Health(); err != nil {
-		return err
+	for _, u := range c.updaters {
```
chbatey: can we log this as well so if an health check fails we see why in the logs?

ingress/frontend.go
```diff
@@ -1,18 +1,10 @@
...
-	// It should block until attaching is complete.
-	Attach() error
-	// Detach should de-register the local proxy from the frontend.
-	// It should block until detaching is complete.
-	Detach() error
-	// Health returns if the number of attached of front ends is equal to the expected
-	// number of front ends
-	Health() error
-}
+type Frontend controller.Updater
```
chbatey: Should we just get rid of this? Or think it is is worth it for documentation? I don't think we ever use it as a Frontend

ingress/proxy.go
```diff
@@ -3,14 +3,4 @@ package ingress
...
-	Start() error
-	// Stop the proxy. Blocks until the proxy stops or an error occurs.
-	Stop() error
-	// Update the proxy configuration. Returns true if the LB was required to reload
-	// its configuration.
-	Update(controller.IngressUpdate) (bool, error)
-	// Health returns nil if healthy, otherwise an error.
-	Health() error
-}
+type Proxy controller.Updater
```
chbatey: Same as updater comment

chbatey:
LGTM with a couple of questions

ingress/frontend.go
jsravn: It's just documentation / package dependency stuff at this point... elb implements Frontend so it doesn't have a dependency on the controller package at all.

jsravn:
I've addressed all the comments, I'll merge once the build goes green.

PR #46, opened 2016-06-13
Store frontend CanonicalHostedZoneNameIDs

balooo:
This information is needed when creating A records in route53 that are
using the TargetAlias functionality.

Needed For: https://github.com/sky-uk/umc-core/issues/772

### DIFF
--- a/elb/elb.go
+++ b/elb/elb.go
@@ -34,8 +34,13 @@ func New(region string, labelValue string, expectedNumber int) controller.Update
 		labelValue:     labelValue,
 		region:         region,
 		expectedNumber: expectedNumber,
-		maxTagQuery:    20,
 	}
+}
+
+// LoadBalancerDetails stores all the elb information we use.
+type LoadBalancerDetails struct {
+	Name         string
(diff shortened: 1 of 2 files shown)


elb/elb.go
```diff
@@ -120,18 +124,26 @@ func (e *elb) Start() error {
 	return nil
 }
 
-func (e *elb) findFrontEndElbs() ([]string, error) {
+// FindFrontEndElbs find all the frontend elbs.
```
chbatey: nit "finds"

Can we add how it does it rather than just repeating the method name as a sentence :)

elb/elb_test.go
```diff
@@ -179,6 +181,26 @@ func TestReportsErrorIfExpectedNotMatched(t *testing.T) {
 	assert.EqualError(t, err, "expected ELBs: 2 actual: 1")
 }
 
+func TestRequiredLoadBalancerDetailsAreExtracted(t *testing.T) {
```
chbatey: Can we name this what it is extracting rather than required?

PR #49, opened 2016-06-14
Remove nginx access log

jsravn:
Until we can prove we need it, let's just remove it. Otherwise it has
unbounded growth.


balooo:
lgtm

PR #50, opened 2016-06-14
Allow annotation replaces default nginx allow

jsravn:
Also add support for comma delimited allows, which is necessary now that
we can't combine with a default allow (such as for external access
combined with internal).


chbatey:
LGTM once build passes

PR #51, opened 2016-06-14
Strip the ingress path prefix

jsravn:
E.g. if someone hits `sky.com/commerce/verification-app/private/ready` it will
proxy to `verification-app/private/ready`, stripping the prefix.

Also fix the location to always have trailing slashes (otherwise it
doesn't work with path prefixes).


nginx/nginx_test.go
```diff
@@ -306,6 +306,38 @@ func TestNginxConfigUpdates(t *testing.T) {
 					ServicePort:    9090,
 					Allow:          "10.82.0.0/16",
 				},
+				{
+					Host:           "chris.com",
```
chbatey: Hrmmmmm

PR #53, opened 2016-06-17
Support kubernetes client certificates

jsravn:
Kubernetes supports private/public keys with client TLS certificates
over token authentication.

### DIFF
--- a/cmd/feed-dns/main.go
+++ b/cmd/feed-dns/main.go
@@ -10,33 +10,46 @@ import (
 )
 
 var (
-	apiServer  string
-	caCertFile string
-	tokenFile  string
-	debug      bool
-	healthPort int
+	apiServer      string
+	caCertFile     string
+	tokenFile      string
+	clientCertFile string
+	clientKeyFile  string
+	debug          bool
+	healthPort     int
 )
 
 func init() {
 	const (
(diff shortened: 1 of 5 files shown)


jsravn:
I was trying to test changes against https://coreos.com/kubernetes/docs/latest/kubernetes-on-vagrant-single.html, but it uses client tls certificates instead of tokens. This adds support to allow you to use client certs, like so:

    docker run -p 8080:8080 -p 8081:8081 -p 12082:12082 -v ~/build/coreos-kubernetes/single-node/ssl:/ssl skycirrus/feed-ingress:50dc60f -apiserver=https://172.17.4.99:443 -cacertfile=/ssl/ca.pem -client-certfile=/ssl/admin.pem -client-keyfile=/ssl/admin-key.pem

chbatey:
Is the reason you would use both? IF not should be barf if both are supplied?

jsravn:
> Is the reason you would use both? IF not should be barf if both are supplied?

Kubernetes supports both if you want. But I don't see why you would want both. Really I added this just to unblock myself testing against local Kubernetes. For our purposes we don't need client certificates, unless we decide to swap from tokens (which are basically a password, and probably less secure) or Kubernetes decides to deprecate tokens or something.

PR #56, opened 2016-06-20
Add upstreams for persistent backend connections

jsravn:
Brings new config:
- Backend keepalive number and keepalive times

And based on internal production nginx:
- Fix nginx keepalive_requests for client requests
- Disable buffering for use with ELB http listeners


jsravn:
This is now ready for review.

balooo:
lgtm

PR #57, opened 2016-06-21
Add nginx prometheus metrics

jsravn:
Add to `/metrics`:

feed_ingress_nginx_connections
feed_ingress_nginx_connections_reading
feed_ingress_nginx_connections_waiting
feed_ingress_nginx_connections_writing
feed_ingress_nginx_accepts
feed_ingress_nginx_handled
feed_ingress_nginx_requests

PR #59, opened 2016-06-22
Track unhealthy time in `/metrics`

jsravn:
So we can alert when the ingress controller is unhealthy.
    
Adds `feed_ingress_unhealthy_time` which is the amount of time in
seconds it has been unhealthy for.

This depends on #57

### DIFF
--- a/util/cmd/cmd.go
+++ b/util/cmd/cmd.go
@@ -49,13 +49,17 @@ func healthHandler(pulse Pulse) func(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-// AddUnhealthyLogger adds a periodic poller which reports an unhealthy status as a log message.
-func AddUnhealthyLogger(pulse Pulse, pollInterval time.Duration) {
+const pollInterval = time.Second
+
+// AddUnhealthyLogger adds a periodic poller which reports an unhealthy status.
+// The healthCounter is increased by pollInterval if unhealthy.
+func AddUnhealthyLogger(pulse Pulse, unhealthyCounter prometheus.Counter) {
 	go func() {
 		healthy := true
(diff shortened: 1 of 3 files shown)


jsravn:
This is passing now, ready for review again.

util/cmd/cmd.go
```diff
@@ -49,13 +49,17 @@ func healthHandler(pulse Pulse) func(w http.ResponseWriter, r *http.Request) {
...
+
+// AddUnhealthyLogger adds a periodic poller which reports an unhealthy status.
+// The healthCounter is increased by pollInterval if unhealthy.
+func AddUnhealthyLogger(pulse Pulse, unhealthyCounter prometheus.Counter) {
 	go func() {
 		healthy := true
 		tickCh := time.Tick(pollInterval)
 		for range tickCh {
 			if err := pulse.Health(); err != nil {
+				unhealthyCounter.Add(pollInterval.Seconds())
```
balooo: why is this increased by pollInterval and not just 1?

balooo:
lgtm, just one question

util/cmd/cmd.go
jsravn: Because it's the time unhealthy. If you just increase by 1, there is no way to know how long it was unhealthy unless you know how often this poller updates (which might slow down under load, etc.)

jsravn: I suppose we could also use it to keep track of uptime statistics.

jsravn:
@balooo happy?

chbatey:
Assigned back to @balooo as i keep looking at this as it has no assignee :)

balooo:
sure

PR #61, opened 2016-06-23
Unhealthy if nginx configuration fails to load

jsravn:
This notifies us if something goes wrong with nginx.conf, which has come
up several times in testing.

If nginx.conf is invalid, it will cause `feed-ingress` to become
unhealthy.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,10 +20,11 @@ RUN apt-key adv --keyserver hkp://pgp.mit.edu:80 --recv-keys 573BFD6B3D8FBC64107
 
 WORKDIR /
 COPY feed-ingress /
+
 RUN mkdir /nginx
-RUN chown nginx:nginx /nginx
+COPY nginx.tmpl /nginx/
+RUN chown -R nginx:nginx /nginx
 
 USER nginx
-COPY nginx.tmpl /nginx/
 
 ENTRYPOINT ["/feed-ingress", "-nginx-workdir", "/nginx"]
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -20,10 +20,11 @@ RUN apt-key adv --keyserver hkp://pgp.mit.edu:80 --recv-keys 573BFD6B3D8FBC64107
 
 WORKDIR /
 COPY feed-ingress /
+
 RUN mkdir /nginx
-RUN chown nginx:nginx /nginx
+COPY nginx.tmpl /nginx/
+RUN chown -R nginx:nginx /nginx
```
jsravn: nginx.tmpl was owned by root before, I changed this so I could edit the template in a running container (to test specifically this PR).

util/cmd/cmd.go
```diff
@@ -39,7 +39,7 @@ func AddHealthPort(pulse Pulse, healthPort int) {
 func healthHandler(pulse Pulse) func(w http.ResponseWriter, r *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
 		if err := pulse.Health(); err != nil {
-			w.WriteHeader(http.StatusServiceUnavailable)
+			w.WriteHeader(http.StatusInternalServerError)
```
jsravn: 503 really doesn't make sense for unhealthy. 503 is meant for temporary downtime during maintenance, etc. In this case we have no idea if the problem is recoverable, and it's probably not intended, so 500 fits better.

jsravn:
This is passing now, ready for review again.

chbatey:
Nice, hadn't realised you could do this

PR #63, opened 2016-06-28
Query load balancers on start up rather than every update

chbatey:


balooo:
Travis still hasn't passed, but tests ran for me.   It would be nice if all the `calculateChanges` tests were actually tested through the public methods, but I don't think it's a big deal.

chbatey:
I can slowly change them but won't do on one big bang

PR #64, opened 2016-06-28
Filter out ingress updates that have an invalid hostname

chbatey:
This also starts to add tests for the R53 Client.

The logs are a little verbose atm I'll cut them down before I finish working in this area.


balooo:
lgtm

PR #65, opened 2016-06-29
Add feed-dns docker release

chbatey:
This adds a build for dns

Hard to test the release phase but tested the docker phase locally


Makefile
```diff
@@ -41,33 +41,47 @@ test :
...
+	  build_files=$${build}_files; \
+	  build_dir=build/$$build; \
+	  binary=$(GOPATH)/bin/feed-$$build; \
+	  docker_file=Dockerfile_$$build; \
+	  mkdir -p $$build_dir; \
+	  cp $$docker_file $${build_dir}/Dockerfile; \
+	  cp $$binary $$build_dir; \
+	done
+
+	cp ${template} ${ingress_build_dir}
```
chbatey: I could have added a var with the extra files in and had an additional for loop but that seemed over kill for one file

Makefile
```diff
@@ -41,33 +41,47 @@ test :
...
 docker : copy
-	@echo "== build docker image"
-	docker build -t $(docker_tag) build/.
-	@echo "Built $(docker_tag)"
+	@echo "== build docker images"
+	@for build in $(builds) ; do \
+	  tag=skycirrus/feed-$$build:$(git_rev) ; \
+	  docker build -t $$tag build/$${build}/. ; \
+	  echo "Built $$tag" ; \
+	done
```
jsravn: I suppose it would be purer to use native makefile "foreach" syntax, rather than using the shell here.

jsravn:
lgtm

Makefile
chbatey: i did look at it, as we have multiple lines to execute it didn't get a little nasty

PR #66, opened 2016-06-29
Handle paging in the call to Route53 GetRecordSets

chbatey:


dns/r53/client.go
```diff
@@ -66,22 +66,34 @@ func (dns *dns) UpdateRecordSets(changes []*route53.Change) error {
...
+		}
 
-	aRecords := []*route53.ResourceRecordSet{}
-	for _, recordSet := range recordSets {
-		if *recordSet.Type == "A" {
-			aRecords = append(aRecords, recordSet)
+		recordSets := recordSetsOutput.ResourceRecordSets
+
+		for _, recordSet := range recordSets {
+			if *recordSet.Type == "A" {
```
jsravn: `"A"`? I'm guessing aws library has a const for this we should probably use instead.

jsravn:
lgtm after addressing comment

PR #67, opened 2016-06-29
Dns large record updates (Depends on #66)

chbatey:
This is based off the #66 branch so do that one first and I will rebase this one

### DIFF
--- a/dns/r53/client.go
+++ b/dns/r53/client.go
@@ -3,10 +3,13 @@ package r53
 import (
 	"fmt"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/route53"
 )
+
+const maxRecordChanges = 100
 
 // Route53Client is the public interface
 type Route53Client interface {
@@ -24,15 +27,17 @@ type r53 interface {
 
 // Route53Client enables interaction with aws route53
 type client struct {
-	r53        r53
-	hostedZone string
+	r53              r53
+	hostedZone       string
+	maxRecordChanges int
 }
 
 // New creates a route53 client used to interact with aws
 func New(region string, hostedZone string) Route53Client {
 	return &client{
-		r53:        route53.New(session.New(), &aws.Config{Region: aws.String(region)}),
-		hostedZone: hostedZone,
+		r53:              route53.New(session.New(), &aws.Config{Region: aws.String(region)}),
+		hostedZone:       hostedZone,
+		maxRecordChanges: maxRecordChanges,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


dns/r53/client.go
```diff
@@ -47,24 +52,35 @@ func (dns *client) GetHostedZoneDomain() (string, error) {
...
 // UpdateRecordSets updates records in aws based on the change list.
-// Todo add tests
 func (dns *client) UpdateRecordSets(changes []*route53.Change) error {
-	recordSetsInput := &route53.ChangeResourceRecordSetsInput{
-		HostedZoneId: aws.String(dns.hostedZone),
-		ChangeBatch: &route53.ChangeBatch{
-			Changes: changes,
-		},
-	}
+	for i := 0; i < len(changes); i += dns.maxRecordChanges {
```
jsravn: Does it need to be qualified with the package name `dns`?

chbatey: Perhaps that is a confusing name for the self reference. It isn't qualified it is a member on the struct so I could override it for testing.

dns/r53/client.go
```diff
@@ -47,24 +52,35 @@ func (dns *client) GetHostedZoneDomain() (string, error) {
...
+		}
 
-	_, err := dns.r53.ChangeResourceRecordSets(recordSetsInput)
+		_, err := dns.r53.ChangeResourceRecordSets(recordSetsInput)
 
-	if err != nil {
-		return fmt.Errorf("failed to create A record: %v", err)
+		if err != nil {
+			return fmt.Errorf("failed to create A record: %v", err)
+		}
```
jsravn: If only we had a generic partition function for slices :).

Do you think it would be clearer what's going on here if we created a partition method which breaks it up into batches first?

dns/r53/client.go
```diff
@@ -47,24 +52,35 @@ func (dns *client) GetHostedZoneDomain() (string, error) {
...
-	recordSetsInput := &route53.ChangeResourceRecordSetsInput{
-		HostedZoneId: aws.String(dns.hostedZone),
-		ChangeBatch: &route53.ChangeBatch{
-			Changes: changes,
-		},
-	}
+	for i := 0; i < len(changes); i += dns.maxRecordChanges {
+		upperBound := min(i+dns.maxRecordChanges, len(changes))
+		batch := changes[i:upperBound]
+		log.Debug("Doing batch ", batch)
```
jsravn: Doesn't it need to be Debugf("Doing batch %v")?

jsravn:
lgtm after comments

dns/r53/client.go
chbatey: That just prints it after

dns/r53/client.go
chbatey: Yes I think so

PR #72, opened 2016-07-01
Add metrics for feed-dns

chbatey:


dns/dns_metrics.go
```diff
@@ -0,0 +1,41 @@
...
+	Name:      "route53_failures",
+	Help:      "The number of failed updates to route53",
+})
+
+var invalidIngressCount = prometheus.NewCounter(prometheus.CounterOpts{
+	Namespace: util.PrometheusNamespace,
+	Subsystem: util.PrometheusDNSSubsystem,
+	Name:      "invalid_ingress_entries",
+	Help:      "The number of invalid ingress entries",
+})
```
jsravn: Good idea... we should probably do this for nginx too.

dns/dns_updater.go
```diff
@@ -129,7 +134,7 @@ func calculateChanges(frontEnds map[string]elb.LoadBalancerDetails,
 	}
 
 	log.Infof("calculated changes to dns: %v", changes)
-	return changes, nil
+	return changes
```
jsravn: Doesn't this disable the healthiness check the controller does for updates?

jsravn:
It's not necessary to start the updater in `main.go`, the generic controller starts all of its updaters.

PR #74, opened 2016-07-06
Don't go unhealthy on long poll disconnects

jsravn:
Watchers have a grace period before they are considered unhealthy.


balooo:
lgtm

PR #75, opened 2016-07-08
Remove test yamls

jsravn:
I'm not sure where these came from. They shouldn't be part of the project
though.


jsravn:
Having an example.yml would be okay though, I think. Showing the annotations and such.

chbatey:
+1 to leaving an example

PR #77, opened 2016-07-11
Fix race conditions in metrics

jsravn:
As detected by the race detector (see
https://travis-ci.org/sky-uk/feed/builds/143821052).


balooo:
lgtm

PR #79, opened 2016-07-12
Add /alive endpoint for liveness probes

jsravn:
Add a very simple 200 endpoint for external liveness probes to make sure
the container is running.


balooo:
lgtm

PR #80, opened 2016-07-12
Add nginx configuration for server_names_hash_*

balooo:
Allows configuration of `server_names_hash_bucket_size` and
`server_names_hash_max_size`

Closes https://github.com/sky-uk/umc-core/issues/1191

### DIFF
--- a/cmd/feed-ingress/main.go
+++ b/cmd/feed-ingress/main.go
@@ -17,56 +17,60 @@ import (
 )
 
 var (
-	debug                        bool
-	apiserverURL                 string
-	caCertFile                   string
-	tokenFile                    string
-	clientCertFile               string
-	clientKeyFile                string
-	ingressPort                  int
-	ingressAllow                 string
-	ingressHealthPort            int
-	healthPort                   int
-	nginxBinary                  string
-	nginxWorkDir                 string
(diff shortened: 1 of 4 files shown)


cmd/feed-ingress/main.go
```diff
@@ -109,6 +113,12 @@ func init() {
 			"times to prevent stale connections.")
 	flag.StringVar(&nginxLogLevel, "nginx-loglevel", defaultNginxLogLevel,
 		"Log level for nginx. See http://nginx.org/en/docs/ngx_core_module.html#error_log for levels.")
+	flag.IntVar(&nginxServerNamesHashBucketSize, "nginx-server-names-hash-bucket-size", defaultNginxServerNamesHashBucketSize,
+		"Sets the bucket size for the server names hash tables.The details of setting up hash tables are provided "+
```
chbatey: Space after the period

chbatey:
Rather than have a default should we only put this in the template if it is passed in?

If this were closed source I wouldn't care as I imagine we'll override it all the time but we're atm wer're defaulting to overriding nginx's server names hash bucket size to 128 where I imagine it'll be 64 on most servers.

chbatey:
Otherwise LGTM

PR #81, opened 2016-07-13
Do not include nginx-server-names-hash-*-size config by default

balooo:


balooo:
Missing file from previous PR

PR #82, opened 2016-07-13
Add apt-get clean before doing an update.

balooo:
To resolve problems caused by apt's local cache not matching upstream

### DIFF
--- a/Dockerfile_dns
+++ b/Dockerfile_dns
@@ -1,9 +1,11 @@
 FROM debian:jessie
 
-RUN apt-get update && apt-get upgrade -y -o Dpkg::Options::="--force-confnew" \
+RUN apt-get clean \
+    && apt-get update && apt-get upgrade -y -o Dpkg::Options::="--force-confnew" \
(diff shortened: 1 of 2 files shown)


Dockerfile_dns
```diff
@@ -15,3 +16,5 @@ RUN apt-get update \
 COPY feed-dns /
 
 ENTRYPOINT ["/feed-dns"]
+
+CMD ["/sbin/my_init"]
```
jsravn: Remove

Dockerfile_dns
```diff
@@ -3,7 +3,8 @@ FROM debian:jessie
 RUN apt-get update && apt-get upgrade -y -o Dpkg::Options::="--force-confnew" \
     && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*  # 2016-06-01
 
-RUN apt-get update \
+RUN apt-get clean \
```
jsravn: Need to do this in line 3 as well

jsravn:
Failed to build the image again - I think we need to clean on the first update step as well.

jsravn:
lgtm

PR #85, opened 2016-07-22
Skip duplicate host/paths

jsravn:
nginx will not load the config if we duplicate locations within a single
server block. This change detects that and skips the duplicate location.

### DIFF
--- a/nginx/nginx.go
+++ b/nginx/nginx.go
@@ -15,6 +15,8 @@ import (
 
 	"syscall"
 
+	"errors"
+
 	log "github.com/Sirupsen/logrus"
 	"github.com/sky-uk/feed/controller"
 	"github.com/sky-uk/feed/util"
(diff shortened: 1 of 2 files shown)


nginx/nginx.go
```diff
@@ -307,21 +309,16 @@ func (lb *nginxLoadBalancer) createConfig(update controller.IngressUpdate) ([]by
 func createNginxEntries(update controller.IngressUpdate) []*nginxEntry {
 	sortedIngressEntries := update.SortedByName().Entries
 	indexedEntries := make(map[string]*nginxEntry)
+	hostPaths := make(map[string]map[string]struct{})
```
chbatey: that is ugly :-/

chbatey: okay i finally groked it, a map of host to map of paths to nothing as go doesn't have sets

chbatey:
lgtm

nginx/nginx.go
jsravn: Maybe we can make it a type to make it more obvious what it is.

chbatey: up to you, i think to the seasoned go programmer it makes sense given the number of things i found about using a map as a set

jsravn:
I made an explicit type to make it more obvious the nested map of maps. I agree it is confusing to read back. Hopefully that makes it more clear.

PR #87, opened 2016-07-29
Disable ingress path stripping by default

jsravn:
Add a flag `-ingress-strip-path` to allow enabling it.


chbatey:
needs rebase

jsravn:
@chbatey rebase is done

PR #91, opened 2016-08-17
Strip path based on annotation and change default to true

chbatey:


chbatey:
I switched the default to true so I can deploy this without making changes to existing apps

cmd/feed-ingress/main.go
```diff
@@ -57,7 +57,7 @@ func init() {
 		defaultIngressPort                    = 8080
 		defaultIngressAllow                   = ""
 		defaultIngressHealthPort              = 8081
-		defaultIngressStripPath               = false
+		defaultIngressStripPath               = true
```
jsravn: Makes sense, to preserve prior behaviour.

controller/controller.go
```diff
@@ -144,14 +149,27 @@ func (c *controller) updateIngresses() error {
...
 					}
 
+					if stripPath, ok := ingress.Annotations[stripPathAnnotation]; ok {
+						if stripPath == "true" {
+							entry.StripPaths = true
+						} else if stripPath == "false" {
+							entry.StripPaths = false
+						} else {
+							log.Debugf("Skipping entry as it has an invalid strip path annotation: %v", stripPath)
+							skipped++
```
jsravn: Should we really skip it? Or would it be better to just assume the default value?

ingress.yml
```diff
@@ -7,6 +7,7 @@ metadata:
   annotations:
     sky.uk/frontend-elb-scheme: internal
     sky.uk/allow: 10.10.82.0/24
+    sky.uk/strip-path: "true"
```
jsravn: don't need double quotes, I think

nginx/nginx_test.go
```diff
@@ -311,6 +310,7 @@ func TestNginxConfigUpdates(t *testing.T) {
 					ServiceAddress: "anotherservice",
 					ServicePort:    6060,
 					Allow:          []string{"10.86.0.0/16"},
+					StripPaths:     false,
```
jsravn: Are we testing the default strip behaviour still?

jsravn:
lgtm apart from a few comments

controller/controller.go
chbatey: We can do that but I will log a warning so we see it

ingress.yml
chbatey: doh

nginx/nginx_test.go
chbatey: this has moved to the controller. It is always set there starting with the default and only overidden of the annotation exists

PR #92, opened 2016-08-24
Allow per ingress backend timeout in nginx

chbatey:
I single timeout is not flexible enough as different services will need
different timeouts.


balooo:
lgtm,  typo in commit `I single timeout` instead of `A single timeout` or `1 single timeout`

chbatey:
thanks

PR #95, opened 2016-09-03
Support PROXY protocol

jsravn:
Add `-nginx-proxy-protocol` which turns on TCP PROXY protocol for the
nginx listeners.


jsravn:
I also refactored `nginx_test.go` a bit, moving duplicate config checks into a single test with test data.

balooo:
lgtm

PR #98, opened 2016-09-15
Disable payload limits

chbatey:
This is required for any request that uploads a file via ingress e.g. a docker
registry


adrianmkng:
lgtm

PR #102, opened 2016-10-05
Toggle access logs via command line argument

sebbonnet:
- Enable access logs when "access-log" command line arg is provided.
- Access logs will be written by default in /var/log/nginx/access.log. The target directory can be overriden via command line arg "access-log-dir"
- Access logs use a custom format to capture various time values between
  nginx and upstream servers
- Access logs are buffered and flushed every minute unless buffer is full

### DIFF
--- a/nginx/nginx.go
+++ b/nginx/nginx.go
@@ -238,15 +238,6 @@ func (lb *nginxLoadBalancer) Update(entries controller.IngressUpdate) error {
 
 func (lb *nginxLoadBalancer) update(entries controller.IngressUpdate) (bool, error) {
 	log.Debugf("Updating loadbalancer %s", entries)
-
-	if lb.Conf.AccessLog {
-		log.Infof("Access log directive enabled. Log files will be written in: %s", lb.Conf.AccessLogDir)
-		if err := os.MkdirAll(lb.Conf.AccessLogDir, 0755); err != nil {
-			log.Errorf("Could not create access log dir. Error: %v", err)
-			return false, err
-		}
-	}
-
 	updatedConfig, err := lb.createConfig(entries)
 	if err != nil {
 		return false, err
--- a/nginx/nginx.go
+++ b/nginx/nginx.go
@@ -43,6 +43,8 @@ type Conf struct {
 	IngressPort                  int
 	LogLevel                     string
 	ProxyProtocol                bool
+	AccessLog                    bool
+	AccessLogDir                 string
 }
 
 // Signaller interface around signalling the loadbalancer process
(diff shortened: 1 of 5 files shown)


sebbonnet:
Sample access logs:
```
10.50.99.45 - - [05/Oct/2016:12:53:25 +0000] "GET / HTTP/1.1" 200 1061 "-" "Monit/5.19.0" "alertmanager.asubdomain.com" uip="10.254.136.189:80" ust="200" rt=0.001 uct="0.000" uht="0.001" urt="0.001"
10.50.96.154 - - [05/Oct/2016:12:53:39 +0000] "PUT /metrics/jobs/feed-ingress/instances/ip-10-50-96-154.eu-west-1.compute.internal HTTP/1.1" 202 0 "-" "Go-http-client/1.1" "pushgateway.asubdomain.com" uip="10.254.128.161:80" ust="202" rt=0.001 uct="0.000" uht="0.001" urt="0.001"
```

nginx/nginx.go
```diff
@@ -236,6 +238,15 @@ func (lb *nginxLoadBalancer) Update(entries controller.IngressUpdate) error {
...
 func (lb *nginxLoadBalancer) update(entries controller.IngressUpdate) (bool, error) {
 	log.Debugf("Updating loadbalancer %s", entries)
+
+	if lb.Conf.AccessLog {
+		log.Infof("Access log directive enabled. Log files will be written in: %s", lb.Conf.AccessLogDir)
+		if err := os.MkdirAll(lb.Conf.AccessLogDir, 0755); err != nil {
+			log.Errorf("Could not create access log dir. Error: %v", err)
+			return false, err
+		}
+	}
```
jsravn: Is this necessary? Can't we just rely on the dockerfile or OS level to create the directory? This also means the feed-ingress process needs write permission in the parent directory.

nginx/nginx.tmpl
```diff
@@ -41,8 +41,15 @@ http {
     real_ip_header {{ if .ProxyProtocol }}proxy_protocol{{ else }}X-Forwarded-For{{ end }};
     real_ip_recursive on;
 
-    # Disable access log
-    access_log             off;
+    # Log format tracking timings
+    log_format upstream_time '$remote_addr - $remote_user [$time_local] '
```
jsravn: Maybe `upstream_info` is better since it's more than just the timings.

nginx/nginx_test.go
```diff
@@ -789,6 +789,26 @@ func TestFailsToUpdateIfConfigurationIsBroken(t *testing.T) {
...
+	lb, _ := newLbWithConf(enabledAccessLogConf)
+
+	assert.NoError(lb.Start())
+	err := lb.Update(controller.IngressUpdate{})
+	assert.NoError(err)
+
+	_, statsErr := os.Stat(tmpDir + "/nginx-access-log")
+	assert.NoError(statsErr)
+}
+
```
jsravn: Missing test coverage that nginx.conf is updated appropriately when the access log is enabled or disabled.

nginx/nginx.go
sebbonnet: We could create the directory at the OS level, if this wasn't a configuration option. Otherwise we would end up with a configuration option that is not used / misleading.
Looking at build-nginx.sh and main.go I can see that we may have that problem already, as sbin-path and config-path are hardcoded in build-nginx.sh, while they can also be provided as a config option in main.go as nginx-binary and nginx-workdir respectively.
So to me the choice is either: we don't make the directory configurable and create it at OS level, or we make it configurable and we create it in the code

nginx/nginx_test.go
sebbonnet: Ok - I'll add the missing coverage

sebbonnet:
@jsravn i've made changes as per review

PR #103, opened 2016-11-14
Introduce iso8601 timestamp and http headers in access logs

mishamo:
https://github.com/sky-uk/umc-core/issues/2121

### DIFF
--- a/nginx/nginx.tmpl
+++ b/nginx/nginx.tmpl
@@ -42,8 +42,8 @@ http {
     real_ip_recursive on;
 
     # Log format tracking timings
-    log_format upstream_info '$remote_addr - $remote_user [$time_local] '
-                             '"$request" $status $body_bytes_sent '
+    log_format upstream_info '$remote_addr - $remote_user [$time_iso8601] '
+                             '"$request" $status{{.AccessLogHeaders}} $body_bytes_sent'
                              '"$http_referer" "$http_user_agent" '
                              '"$host" uip="$upstream_addr" ust="$upstream_status" '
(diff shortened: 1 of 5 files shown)


nginx/nginx.tmpl
```diff
@@ -42,8 +42,8 @@ http {
     real_ip_recursive on;
 
     # Log format tracking timings
-    log_format upstream_info '$remote_addr - $remote_user [$time_local] '
-                             '"$request" $status $body_bytes_sent '
+    log_format upstream_info '$remote_addr - $remote_user [$time_iso8601] '
+                             '"$request" $status{{.AccessLogHeaders}} $body_bytes_sent'
```
jsravn: Doesn't it need a space between `$status` and `{{.AccessLogHeaders}}`?

nginx/nginx_test.go
```diff
@@ -268,6 +271,14 @@ func TestNginxConfig(t *testing.T) {
 				"access_log /nginx-access-log/access.log upstream_info buffer=32k flush=1m;",
 			},
 		},
+		{
+			"Access logs use custom headers when enabled",
+			logHeadersConf,
+			[]string{
+				"$http_Content-Type",
+				"$http_Authorization",
```
jsravn: Can we check the entire line, so we make sure it's part of the access log?

So
```golang
[]string {
  "'$request $status $http_content-Type $http_Authorization $body_bytes_sent",
}
```

nginx/nginx.tmpl
mishamo: Shouldn't do, each iteration of prepending $http_ to each header also prepends a space

PR #106, opened 2016-11-25
Basic throttling for nginx updates

chbatey:
If many updates came in in a short space of time nginx spawns
many workers processes resulting in OOM kills

### DIFF
--- a/cmd/feed-ingress/main.go
+++ b/cmd/feed-ingress/main.go
@@ -17,39 +17,26 @@ import (
 )
 
 var (
-	debug                             bool
-	apiserverURL                      string
-	caCertFile                        string
-	tokenFile                         string
-	clientCertFile                    string
-	clientKeyFile                     string
-	ingressPort                       int
-	ingressAllow                      string
-	ingressHealthPort                 int
-	ingressStripPath                  bool
-	healthPort                        int
-	nginxBinary                       string
-	nginxWorkDir                      string
-	nginxWorkerProcesses              int
-	nginxWorkerConnections            int
-	nginxKeepAliveSeconds             int
-	nginxBackendKeepalives            int
-	nginxBackendKeepaliveSeconds      int
-	nginxBackendConnectTimeoutSeconds int
-	nginxLogLevel                     string
-	nginxTrustedFrontends             string
-	nginxServerNamesHashBucketSize    int
-	nginxServerNamesHashMaxSize       int
-	nginxProxyProtocol                bool
-	elbLabelValue                     string
-	elbRegion                         string
-	elbExpectedNumber                 int
-	pushgatewayURL                    string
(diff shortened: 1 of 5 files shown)


cmd/feed-ingress/main.go
```diff
@@ -17,39 +17,26 @@ import (
...
-	elbLabelValue                     string
-	elbRegion                         string
-	elbExpectedNumber                 int
-	pushgatewayURL                    string
-	pushgatewayIntervalSeconds        int
-	pushgatewayLabels                 cmd.KeyValues
-	accessLog                         bool
-	accessLogDir                      string
-	nginxLogHeaders                   string
+	debug                      bool
```
jsravn: Nice

nginx/nginx.go
```diff
@@ -202,6 +231,23 @@ func (lb *nginxLoadBalancer) periodicallyUpdateMetrics() {
 	}
 }
 
+func (lb *nginxLoadBalancer) signaler() {
```
jsravn: Can we improve the function name, signaller is the thing that signals nginx and we already use that name. This is the loop.

nginx/nginx.go
```diff
@@ -70,15 +73,36 @@ func (s *osSignaller) sighup(p *os.Process) error {
...
+	lb.updateRequired.Lock()
+	defer lb.updateRequired.Unlock()
+	lb.updateRequired.update = true
+}
+
+func (lb *nginxLoadBalancer) signalIfRequired() {
+	lb.updateRequired.Lock()
+	defer lb.updateRequired.Unlock()
+	if lb.updateRequired.update {
+		lb.signaller.sighup(lb.cmd.Process)
```
jsravn: Can we log when it gets updated?

nginx/nginx.go
```diff
@@ -221,25 +267,22 @@ func (lb *nginxLoadBalancer) Stop() error {
...
 	}
 
 	if updated {
-		err = lb.signaller.sighup(lb.cmd.Process)
-		if err != nil {
-			return fmt.Errorf("unable to signal nginx to reload: %v", err)
-		}
-		log.Info("Nginx updated")
+		lb.signalRequired()
+		log.Infof("Nginx update required (can take up to %d seconds)", lb.Conf.UpdateFrequencySeconds)
```
jsravn: Not sure this is worthwhile logging - if we put a log message when it does the actual update. We'll already get lots of log spam when an update comes in.

nginx/nginx.go
chbatey: done

nginx/nginx.go
chbatey: yep

nginx/nginx.go
chbatey: removed

PR #107, opened 2016-11-30
Synchronously update nginx for the first update

chbatey:
Otherwise ingress can be unavailble for up to 30s

### DIFF
--- a/elb/elb.go
+++ b/elb/elb.go
@@ -8,6 +8,8 @@ import (
 
 	"errors"
 
+	"sync"
+
 	log "github.com/Sirupsen/logrus"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
@@ -31,6 +33,7 @@ func New(region string, labelValue string, expectedNumber int) controller.Update
 		labelValue:     labelValue,
 		region:         region,
 		expectedNumber: expectedNumber,
+		initialised:    initialised{},
 	}
 }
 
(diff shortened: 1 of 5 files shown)


nginx/nginx.go
```diff
@@ -275,7 +275,11 @@ func (lb *nginxLoadBalancer) Update(entries controller.IngressUpdate) error {
 		return fmt.Errorf("unable to update nginx: %v", err)
 	}
 
-	if updated {
+	if updated && !lb.initialUpdateApplied.Get() {
+		log.Info("Initial update. Signalling nginx synchronously.")
+		lb.signaller.sighup(lb.cmd.Process)
+		lb.initialUpdateApplied.Set(true)
+	} else if updated {
```
jsravn: Any reason to not nest the conditional? It's a little confusing as-is I think.

```go
if updated {
  if !lb.initialUpdateApplied.Get() {
     updateSync()
  } else {
     updateAsync()
  }
}
```

chbatey: can do

chbatey:
Hrmm I agree. Q reasonable change given "updaters" are currently independent

chbatey:
Pushed a second commit @jsravn that puts the register ELBs in the first update. I didn't use `sync.Once` as that doesn't appear to allow me to return a value which makes sense.

elb/elb.go
```diff
@@ -51,6 +54,12 @@ type elb struct {
 	instanceID          string
 	elbs                map[string]LoadBalancerDetails
 	registeredFrontends int
+	initialised         initialised
+}
+
+type initialised struct {
+	sync.Mutex
+	done bool
 }
```
jsravn: You can use the atomic bool inside the util package.

PR #108, opened 2016-12-05
Swap to client-go

jsravn:
Vendoring is split up into its own commit so the changes can be reviewed separately.


cmd/feed-dns/main.go
```diff
@@ -94,10 +88,8 @@ func main() {
 	cmd.AddHealthPort(controller, healthPort)
 	cmd.AddSignalHandler(controller)
 
-	err := controller.Start()
-	if err != nil {
-		log.Error("Error while starting controller: ", err)
-		os.Exit(-1)
+	if err := controller.Start(); err != nil {
+		log.Fatal("Error while starting controller: ", err)
```
balooo: Why do we not have to exit here?

PR #109, opened 2016-12-06
Add logrotate for nginx logs

chbatey:


scripts/logrotate.config
```diff
@@ -0,0 +1,11 @@
+/var/log/nginx/*.log {
+        size 400M
+        rotate 10
+        notifempty
+        missingok
+        create 640 nginx nginx
+        compress
+        postrotate
```
totahuanocotl: does this need to be run only once, or once for each log which is rotated?
if it should be only once should it be using the `sharedscripts` directive?

chbatey: i don't know i just took it from umc-core feed :)

sebbonnet: In umc-core feed we used to have only only access logs, so the USR1 directive would only ever be invoked once on access.log rotation.
Given that this is meant to be generic, we may have multiple type of logs, so `sharedscrip` would be best

PR #113, opened 2016-12-21
Fix backend keepalive seconds

jsravn:
Don't use the client keepalive, instead use what is specified on the
command line.

### DIFF
--- a/cmd/feed-ingress/main.go
+++ b/cmd/feed-ingress/main.go
@@ -19,23 +19,24 @@ import (
 )
 
 var (
-	debug                      bool
-	kubeconfig                 string
-	resyncPeriod               time.Duration
-	ingressPort                int
-	ingressAllow               string
-	ingressHealthPort          int
-	ingressStripPath           bool
-	healthPort                 int
-	elbLabelValue              string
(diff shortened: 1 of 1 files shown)


cmd/feed-ingress/main.go
```diff
@@ -165,7 +166,7 @@ func main() {
 		Updaters:                updaters,
 		DefaultAllow:            ingressAllow,
 		DefaultStripPath:        ingressStripPath,
-		DefaultBackendKeepAlive: nginxConfig.KeepaliveSeconds,
+		DefaultBackendKeepAlive: nginxBackendKeepaliveSeconds,
```
chbatey: Given we just copy this in without modifying it should we just write directly into the struct like with other nginx config?

jsravn: Pre initialise the controller config struct you mean? Just feels so dirty to me :)

jsravn: Ok I've made that change. Wasn't that bad.

jsravn:
Its hiding my comment but I made the change requested.

PR #119, opened 2017-01-13
Improve upstream names and organisation

jsravn:
- Names now reflect the backend service
- Common upstreams are now shared rather than being duplicated

### DIFF
--- a/nginx/nginx.go
+++ b/nginx/nginx.go
@@ -353,7 +353,7 @@ func (n *nginxUpdater) getNginxLogHeaders() string {
 type pathSet map[string]struct{}
 
 func createNginxEntries(update controller.IngressUpdate) []*nginxEntry {
-	sortedIngressEntries := update.SortedByName().Entries
+	sortedIngressEntries := update.SortedByNamespaceName().Entries
 	hostToNginxEntry := make(map[string]*nginxEntry)
 	hostToPaths := make(map[string]pathSet)
 	var nginxEntries []*nginxEntry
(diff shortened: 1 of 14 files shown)


jsravn:
We refactored nginx.go quite a lot as well in the first two commits, which is why they are separate commits.

jsravn:
Had to update travis build to use python3.6, as 3.5 has issues I think with signals and `time.sleep`.

nginx/nginx.go
```diff
@@ -349,60 +350,99 @@ func (lb *nginxLoadBalancer) createConfig(update controller.IngressUpdate) ([]by
...
+	return fmt.Sprintf("%s.%s.%d", e.Namespace, e.ServiceAddress, e.ServicePort)
+}
+
+type servers []*server
+
+func (s servers) Len() int           { return len(s) }
+func (s servers) Less(i, j int) bool { return s[i].ServerName < s[j].ServerName }
+func (s servers) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+
+type locations []location
```
chbatey: Any reason for servers being a pointer and this a value?

nginx/nginx.go
```diff
@@ -349,60 +350,99 @@ func (lb *nginxLoadBalancer) createConfig(update controller.IngressUpdate) ([]by
...
+		serverEntry.Name += " " + ingressEntry.NamespaceName()
+		serverEntry.Locations = append(serverEntry.Locations, location)
+	}
 
-		entry.Name += " " + ingressEntry.Name
-		entry.Upstreams = append(entry.Upstreams, upstream)
-		entry.Locations = append(entry.Locations, location)
-		upstreamIndex++
+	var serverEntries []*server
+	for _, serverEntry := range hostToNginxEntry {
```
chbatey: is the extra sorting to avoid an update when nothing has changed and go gives it back in a different order?

jsravn:
> Is there any character that can be in an ingress name that can't be in an nginx conf file?

@chbatey Not that we could tell. We did test this. Ingress names are restricted to the same requirements as a dns name, so we tried all possible things in a dns name (-, numbers), by updating nginx.conf manually and reloading.

nginx/nginx.go
jsravn: Yeah it was done that way because in `#createServerEntries` we need a pointer to the server struct, as we mutate its state as we iterate through the ingress entries. I agree it's a bit confusing - we could change locations to be a pointer as well.

If we make servers not a slice-to-pointer, then we'll have to write code to convert the slice by copying over all its contents, which I think is a bit crap!

nginx/nginx.go
jsravn: Yes, we were trying to maintain a stable ordering on every update.

jsravn:
I changed the locations to be a slice of pointers.

PR #123, opened 2017-01-18
Use HostedZoneID not DNSName for LB hosted zone

jsravn:


jsravn:
Tests were converted to ALBs, as the ELB stuff will be removed shortly. Unfortunately that led to a regression in the ELB code! I don't think it's worth adding the test coverage now as we'll remove it very soon.

PR #124, opened 2017-01-23
Use endpoints

jsravn:


nginx/nginx.tmpl
```diff
@@ -89,11 +89,20 @@ http {
...
 {{- range $upstream := .Upstreams }}
     upstream {{ $upstream.ID }} {
-        server {{ $upstream.Server }};
+{{- range $address := $upstream.Addresses }}
+        server {{ $address }}:{{ $upstream.Port }};
+{{ end }}
         keepalive {{ $keepalive }};
+        # Keep connections balanced - especially useful for rolling deployments.
+        least_conn;
+        # Health checks - check every 5 seconds, expect new endpoints to be down by default.
```
totahuanocotl: this comment does not match the actual configuration

nginx/nginx.tmpl
```diff
@@ -89,11 +89,20 @@ http {
...
     
 {{- range $upstream := .Upstreams }}
     upstream {{ $upstream.ID }} {
-        server {{ $upstream.Server }};
+{{- range $address := $upstream.Addresses }}
+        server {{ $address }}:{{ $upstream.Port }};
+{{ end }}
         keepalive {{ $keepalive }};
+        # Keep connections balanced - especially useful for rolling deployments.
+        least_conn;
```
sebbonnet: Out of curiosity - did you consider using `least_time`?

nginx/nginx.tmpl
```diff
@@ -167,6 +176,11 @@ http {
             vhost_traffic_status_display_format html;
         }
 
+        location /upstreams {
```
sebbonnet: what does this do?

nginx/nginx.tmpl
jsravn: least_time is only in the commercial nginx.

nginx/nginx.tmpl
jsravn: Adds an endpoint to show the health check stats on upstreams.

controller/ingress_entry.go
```diff
@@ -35,21 +42,13 @@ type IngressEntry struct {
...
-	}
-	if e.ServicePort == 0 {
-		return errors.New("missing service port")
-	}
-	return nil
-}
-
 // NamespaceName returns the string "Namespace/Name".
-func (e IngressEntry) NamespaceName() string {
+func (e *IngressEntry) NamespaceName() string {
```
jsravn: I'm trying to avoid unnecessary copying of the struct.

controller/controller.go
```diff
@@ -207,17 +207,52 @@ func (c *controller) updateIngresses() error {
...
-func mapNamesToAddresses(services []*v1.Service) map[serviceName]string {
-	m := make(map[serviceName]string)
+func mapEndpointsToServices(multipleEndpoints []*v1.Endpoints) map[serviceKey]Service {
+	m := make(map[serviceKey]Service)
+
+	for _, endpoints := range multipleEndpoints {
+		for _, subset := range endpoints.Subsets {
+			for _, port := range subset.Ports {
+
+				if port.Protocol == v1.ProtocolUDP {
```
jsravn: Yep protocol is optional and can be blank. So we just want the explicit check here.

controller/controller.go
```diff
@@ -128,66 +125,69 @@ func (c *controller) updateIngresses() error {
...
-					skipped = append(skipped, fmt.Sprintf("%s/%s (service doesn't exist)", ingress.Namespace, ingress.Name))
 				}
+
+				if backendKeepAlive, ok := ingress.Annotations[backendKeepAliveSeconds]; ok {
+					tmp, _ := strconv.Atoi(backendKeepAlive)
+					entry.BackendKeepAliveSeconds = tmp
+				}
+
+				if err := validate(&entry); err != nil {
+					skipped = append(skipped, fmt.Sprintf("%s (%v)", entry.NamespaceName(), err))
```
jsravn: This is already the ingress name.

controller/controller.go
```diff
@@ -128,66 +125,69 @@ func (c *controller) updateIngresses() error {
...
+					ELbScheme:               ingress.Annotations[frontendElbSchemeAnnotation],
+					StripPaths:              c.defaultStripPath,
+					BackendKeepAliveSeconds: c.defaultBackendTimeout,
+				}
 
-					if backendKeepAlive, ok := ingress.Annotations[backendKeepAliveSeconds]; ok {
-						tmp, _ := strconv.Atoi(backendKeepAlive)
-						entry.BackendKeepAliveSeconds = tmp
+				if allow, ok := ingress.Annotations[ingressAllowAnnotation]; ok {
+					if allow == "" {
```
jsravn: Nope, it returns a single element array with `""`.

controller/controller_test.go
```diff
@@ -386,73 +380,105 @@ func TestUpdaterIsUpdatedOnK8sUpdates(t *testing.T) {
...
-		serviceCh <- struct{}{}
+		endpointsCh <- struct{}{}
 		time.Sleep(smallWaitTime)
 
 		//then
 		assert.NoError(controller.Stop())
 		time.Sleep(smallWaitTime)
 		updater.AssertExpectations(t)
+
+		if t.Failed() {
```
jsravn: It won't fail until all the examples finish in the test function. This makes it much easier to see which example failed.

jsravn:
I'm going to remove the health checks as well - I think they are redundant to kubelet's probes, and it will require more code/configuration that duplicates the probes.

controller/ingress_entry.go
sebbonnet: then we should do the same for String() method

jsravn:
Please see my last commit - I left old one alone so it's easier to review.

jsravn:
I also added proxy_next_upstream and max_fail/fail_timeout configuration, which we missed in the original commit. These are necessary to control how nginx deals with failures when there are multiple endpoints.

nginx/nginx.tmpl
```diff
@@ -87,22 +87,21 @@ http {
 
     # Start ingresses
     {{- $port := .IngressPort }}
-    {{- $keepalive := .BackendKeepalives }}
+    {{- $keepalive := .UpstreamKeepalives }}
     {{- $proxyprotocol := .ProxyProtocol }}
-    {{- $connectTimeout := .BackendConnectTimeoutSeconds }}
+    {{- $connectTimeout := .ProxyConnectTimeoutSeconds }}
```
sebbonnet: connectTimeout is not used anymore

PR #128, opened 2017-02-09
Don't log 404s on error level

jsravn:
To prevent spam when using an error log level of 'error'.


sebbonnet:
@jsravn broken build

jsravn:
Looks like some apt brokenness, restarting build.

PR #134, opened 2017-03-11
Deterministic ordering for duplicate ingresses

sebbonnet:
Keep the most recent ingress when duplicate ingresses found to make the order
deterministic, so that new feed deployment or reload always keep the same
ingress when duplicates are found.

The inconsistent ordering originates from`Client#GetIngresses` which returns ingresses in a different order even when they do not change. 

Fix sky-uk/umc-core#2764

### DIFF
--- a/nginx/nginx.go
+++ b/nginx/nginx.go
@@ -406,31 +406,21 @@ type pathSet map[string]bool
 
 func createServerEntries(update controller.IngressUpdate) []*server {
 	hostToNginxEntry := make(map[string]*server)
-	hostToPaths := make(map[string]pathSet)
 
-	for _, ingressEntry := range update.Entries {
+	for _, ingressEntry := range uniqueIngressEntries(update.Entries) {
 		serverEntry, exists := hostToNginxEntry[ingressEntry.Host]
 		if !exists {
 			serverEntry = &server{ServerName: ingressEntry.Host}
 			hostToNginxEntry[ingressEntry.Host] = serverEntry
-			hostToPaths[ingressEntry.Host] = make(map[string]bool)
 		}
 
-		nginxPath := createNginxPath(ingressEntry.Path)
 		location := location{
-			Path:                    nginxPath,
+			Path:                    createNginxPath(ingressEntry.Path),
 			UpstreamID:              upstreamID(ingressEntry),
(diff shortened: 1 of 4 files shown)


nginx/nginx.go
```diff
@@ -448,6 +438,42 @@ func createServerEntries(update controller.IngressUpdate) []*server {
...
+	}
+
+	uniqueIngressEntries := make([]controller.IngressEntry, 0, len(uniqueIngress))
+	for _, value := range uniqueIngress {
+		uniqueIngressEntries = append(uniqueIngressEntries, value)
+	}
+	return uniqueIngressEntries
+}
+
+func orderByCreation(entry1, entry2 controller.IngressEntry) (mostRecentIngress, olderIngress controller.IngressEntry) {
```
jsravn: For future reference don't use named return parameters, it's considered bad practice in golang.

jsravn: I guess you could argue this method is short enough, but naked returns are pretty rare in golang code I would avoid using them except for the simplest method that is a line or two.

nginx/nginx.go
```diff
@@ -448,6 +438,42 @@ func createServerEntries(update controller.IngressUpdate) []*server {
...
+		log.Infof("Ignoring '%s' because it duplicates the host/path of a more recent entry '%s'", olderIngress.NamespaceName(), mostRecentIngress.NamespaceName())
+		uniqueIngress[key] = mostRecentIngress
+	}
+
+	uniqueIngressEntries := make([]controller.IngressEntry, 0, len(uniqueIngress))
+	for _, value := range uniqueIngress {
+		uniqueIngressEntries = append(uniqueIngressEntries, value)
+	}
+	return uniqueIngressEntries
+}
```
jsravn: This method could be a lot simpler I think if we simply sorted and then discarded duplicate keys. No need for a map either.

nginx/nginx.go
jsravn: Re-reading https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters , they seem a bit more lax on this than I would be :). I find it hard to understand when the return variable can be assigned anywhere in the method.

PR #135, opened 2017-03-11
Accept env variables to define version and repo

sebbonnet:
To make it easier to test a change before releasing by allowing to specify build version / target
respository to use

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,16 @@ test :
 # Docker build 
 ingress_build_dir := build/ingress
 template := ./nginx/nginx.tmpl
-git_rev := $(shell git rev-parse --short HEAD)
+ifdef REPOSITORY
+    repository := $(REPOSITORY)
+else 
+    repository := skycirrus
+endif
+ifdef BUILD_VERSION
+    build_version := $(BUILD_VERSION)
+else 
+    build_version := $(shell git rev-parse --short HEAD)
+endif
 builds := ingress dns
 
 clean :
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -43,7 +43,16 @@ test :
 # Docker build 
 ingress_build_dir := build/ingress
 template := ./nginx/nginx.tmpl
-git_rev := $(shell git rev-parse --short HEAD)
+ifdef REPOSITORY
+    repository := $(REPOSITORY)
+else 
+    repository := skycirrus
+endif
```
jsravn: You don't need ifdefs, just do:

```
REPOSITORY=skycirrus
```

Then use that everywhere. If someone then does `make REPOSITORY=mytest` it will override it.

Same with BUILD_VERSION below.

Makefile
```diff
@@ -79,9 +88,22 @@ release : docker
...
+	@for build in $(builds) ; do \
+	  set -e; \
+	  image_tag=${repository}/feed-$$build:${build_version} ; \
+	  test_tag=${repository}/test/feed-$$build:${build_version} ; \
+	  latest_tag=${repository}/test/feed-$$build:latest ; \
+	  docker tag $$image_tag $$test_tag ; \
+	  docker tag $$test_tag $$latest_tag ; \
+	  docker push $$test_tag ; \
+	  docker push $$latest_tag ; \
+	done
```
jsravn: I'm not convinced we need or want this. Why can't you just manually do a docker push if you want to test it?

We could also add a test repo for building the branches and do that instead of allowing overrides of the variables.

Makefile
sebbonnet: ok

Makefile
sebbonnet: We need a test repository that does not require login in - the problem with dockerhub is that you need to login in order to push any image.  
Besides you have to tag the image with the correct repo name before it can be pushed it, and of course before you do that you need to build it. That's 3 commands. So `releasetotest`  is simply a mechanism to do all that in one command without having to log in.

sebbonnet:
@jsravn I've created #137 instead of adding the `releasetotest` target.
I've kept the REPOSITORY and BUILD_VERSION variables to make it possible to overwrite them as you suggested

PR #136, opened 2017-03-14
Update route53 recordsets only when necessary

sebbonnet:
Send route53 changerecordset requests when hostname or ELB change.
This was done to avoid issuing route53 requests update when they are no changes, as these may fail.
Also append a "." to the alb dns name in all cases for consistency.

### DIFF
--- a/dns/dns_updater.go
+++ b/dns/dns_updater.go
@@ -118,7 +118,7 @@ func (u *updater) initALBs() error {
 		}
 
 		for _, lb := range resp.LoadBalancers {
-			u.schemeToDNS[*lb.Scheme] = dnsDetails{dnsName: *lb.DNSName, hostedZoneID: *lb.CanonicalHostedZoneId}
+			u.schemeToDNS[*lb.Scheme] = dnsDetails{dnsName: *lb.DNSName + ".", hostedZoneID: *lb.CanonicalHostedZoneId}
 		}
 
 		if resp.NextMarker == nil {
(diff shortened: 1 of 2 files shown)


dns/dns_updater.go
```diff
@@ -242,6 +246,15 @@ func (u *updater) createChanges(hostToIngress hostToIngress,
...
 }
 
+func findMatchingRecord(host string, originalRecords []*route53.ResourceRecordSet) (bool, *route53.ResourceRecordSet) {
+	for _, recordSet := range originalRecords {
+		if host == *recordSet.Name {
+			return true, recordSet
+		}
+	}
+	return false, nil
+}
```
jsravn: Do you think it would be better if we used a map of the original records instead of having to iterate through them each time? We could create a map with key of host + elb DNS name. I think it would simplify the logic a bit too.

sebbonnet: good idea

sebbonnet: actually just a map keyed by host would be better

dns/dns_updater.go
```diff
@@ -226,7 +230,11 @@ func (u *updater) createChanges(hostToIngress hostToIngress,
 			continue
 		}
 
-		changes = append(changes, newChange("UPSERT", host, dnsDetails.dnsName, dnsDetails.hostedZoneID))
+		matchingRecordSet, recordExists := hostToRecords[host]
+		elbHasChanged := recordExists && dnsDetails.dnsName != *matchingRecordSet.AliasTarget.DNSName
```
jsravn: If you make hostToRecord have the key `{host, elbDNSName}` you don't need any of this logic.

dns/dns_updater.go
```diff
@@ -226,7 +231,10 @@ func (u *updater) createChanges(hostToIngress hostToIngress,
 			continue
 		}
 
-		changes = append(changes, newChange("UPSERT", host, dnsDetails.dnsName, dnsDetails.hostedZoneID))
+		_, recordExists := indexedRecords[recordKey{host, dnsDetails.dnsName}]
+		if !recordExists {
```
jsravn: You can inline the variable.

PR #141, opened 2017-06-16
Feed reporting unhealthy at startup

alangibson01:
Short summary: until such a time as the first ELB or ALB update attempt has happened, don't report the updater as being unhealthy.
I've also added a fix to the ELB updater, as up until now its healthcheck has always returned healthy, regardless of whether the update operation actually succeeded or not. It now functions in the same way as the ALB healthcheck.

Fixes https://github.com/sky-uk/umc-core/issues/3447


jsravn:
Fixes https://github.com/sky-uk/umc-core/issues/3447

jsravn:
@alangibson01 Can you fix the commit author? Looks like you're using the old cirrus-ps01 account.

alb/alb.go
```diff
@@ -106,7 +110,7 @@ func (a *alb) Stop() error {
 
 // Health returns nil if attached to all frontends.
 func (a *alb) Health() error {
-	if len(a.targetGroupNames) == a.registeredFrontends {
+	if !a.readyForHealthCheck.Get() || len(a.targetGroupNames) == a.registeredFrontends {
```
jsravn: I believe the usage of registeredFrontends needs to be guarded by a mutex. Since `#Health` will be called in a separate goroutine from `#Update`. `targetGroupNames` is okay because it is initialised before the goroutines are started.

elb/elb.go
```diff
@@ -225,18 +226,25 @@ func (e *elb) Stop() error {
 	return nil
 }
 
-// Health is a no-op atm. Could be updated in the future to check all the ELBs are still attached
 func (e *elb) Health() error {
-	return nil
+	if !e.readyForHealthCheck.Get() || e.expectedNumber == e.registeredFrontends {
```
jsravn: Same as alb.go, registeredFrontends needs to be guarded by a mutex.

elb/elb_test.go
```diff
@@ -448,3 +448,85 @@ func TestDeRegisterInstanceError(t *testing.T) {
...
+	e, mockElb, mockMetadata := setup()
+	instanceID := "cow"
+	mockInstanceMetadata(mockMetadata, instanceID)
+	clusterFrontEnd := "cluster-frontend"
+	mockLoadBalancers(mockElb,
+		lb{name: clusterFrontEnd, scheme: elbInternalScheme})
+	mockClusterTags(mockElb,
+		lbTags{name: clusterFrontEnd, tags: []*aws_elb.Tag{{Key: aws.String(frontendTag), Value: aws.String(clusterName)}}},
+	)
+	mockElb.On("RegisterInstancesWithLoadBalancer", mock.Anything).Return(&aws_elb.RegisterInstancesWithLoadBalancerOutput{}, errors.New("no register for you"))
```
jsravn: Can you break these long lines up, we try to keep < 120 chars.

elb/elb_test.go
```diff
@@ -448,3 +448,85 @@ func TestDeRegisterInstanceError(t *testing.T) {
...
+
+	// when
+	err := e.Start()
+	updateErr := e.Update(controller.IngressUpdate{})
+
+	// then
+	assert.NoError(t, err)
+	assert.NoError(t, updateErr)
+	assert.NoError(t, e.Health())
+}
```
jsravn: I think we already do this test, in `TestAttachWithSingleMatchingLoadBalancers`.

alb/alb_test.go
```diff
@@ -298,3 +298,54 @@ func TestStopWaitsForDeregisterDelay(t *testing.T) {
...
+
+	//when
+	err := a.Start()
+	updateErr := a.Update(controller.IngressUpdate{})
+
+	//then
+	assert.NoError(t, err)
+	assert.NoError(t, updateErr)
+	assert.Nil(t, a.Health())
+}
```
jsravn: I think we could just add `assert.Nil(t, a.Health())` to the preexisting test `TestRegisterInstance`.

elb/elb_test.go
alangibson01: Done

elb/elb_test.go
alangibson01: Good spot.

alb/alb_test.go
alangibson01: Done

alb/alb.go
alangibson01: Done

elb/elb.go
alangibson01: Done

jsravn:
Released as v1.0.1.

PR #143, opened 2017-06-27
Start Nginx only after first update is received

alangibson01:
Prior to this change, there was a small window where the Nginx instance
used by feed-ingress could be active but without having loaded its
configuration. In this instance, the load balancer would be able to
reach it, but would receive a 404 response.
    
The solution is to only start Nginx after the first update has been
received and the configuration is ready - this means the load balancer
will only be able to reach it when it is suitably configured.
    
Fixes https://github.com/sky-uk/umc-core/issues/3586

### DIFF
--- a/nginx/nginx.go
+++ b/nginx/nginx.go
@@ -132,7 +132,7 @@ func (c *Conf) nginxConfFile() string {
 }
 
 // New creates an nginx updater.
-func New(nginxConf Conf) controller.Updater {
+func New(nginxConf Conf) *nginxUpdater {
 	initMetrics()
 
 	nginxConf.WorkingDir = strings.TrimSuffix(nginxConf.WorkingDir, "/")
@@ -189,7 +189,7 @@ func (n *nginxUpdater) ensureNginxRunning() error {
 	defer n.initialUpdateApplied.Unlock()
 
 	if !n.initialUpdateApplied.done {
-		log.Info("Starting nginx with initial configuration")
+		log.Info("Starting nginx for the first time")
 		if err := n.nginx.Start(); err != nil {
 			return fmt.Errorf("unable to start nginx: %v", err)
 		}
@@ -506,6 +506,12 @@ func createNginxPath(rawPath string) string {
 }
 
 func (n *nginxUpdater) Health() error {
+	n.initialUpdateApplied.Lock()
+	defer n.initialUpdateApplied.Unlock()
+
+	if !n.initialUpdateApplied.done {
+		return errors.New("waiting for initial update")
+	}
 	if !n.running.Get() {
 		return errors.New("nginx is not running")
 	}
--- a/nginx/nginx.go
+++ b/nginx/nginx.go
@@ -89,14 +89,14 @@ type nginxUpdater struct {
 	running                util.SafeBool
 	lastErr                util.SafeError
 	metricsUnhealthy       util.SafeBool
-	initialUpdateApplied   initialUpdateApplied
+	nginxStarted           nginxStarted
 	initialUpdateAttempted util.SafeBool
 	doneCh                 chan struct{}
 	nginx                  *nginx
 	updateRequired         util.SafeBool
 }
 
-type initialUpdateApplied struct {
+type nginxStarted struct {
 	sync.Mutex
 	done bool
 }
@@ -186,10 +186,10 @@ func (n *nginxUpdater) initialiseNginxConf() error {
 func (n *nginxUpdater) ensureNginxRunning() error {
 	// Guard against the (hopefully unlikely) event that two updates could be
 	// received concurrently and attempt to start Nginx twice.
-	n.initialUpdateApplied.Lock()
-	defer n.initialUpdateApplied.Unlock()
+	n.nginxStarted.Lock()
+	defer n.nginxStarted.Unlock()
 
-	if !n.initialUpdateApplied.done {
+	if !n.nginxStarted.done {
 		log.Info("Starting nginx for the first time")
 		if err := n.nginx.Start(); err != nil {
 			return fmt.Errorf("unable to start nginx: %v", err)
@@ -206,7 +206,7 @@ func (n *nginxUpdater) ensureNginxRunning() error {
 		go n.periodicallyUpdateMetrics()
 		go n.backgroundSignaller()
 
-		n.initialUpdateApplied.done = true
+		n.nginxStarted.done = true
 	}
 
 	return nil
@@ -508,9 +508,6 @@ func createNginxPath(rawPath string) string {
 }
 
 func (n *nginxUpdater) Health() error {
-	n.initialUpdateApplied.Lock()
-	defer n.initialUpdateApplied.Unlock()
-
 	if !n.initialUpdateAttempted.Get() {
 		return errors.New("waiting for initial update")
 	}
--- a/nginx/nginx.go
+++ b/nginx/nginx.go
@@ -132,7 +132,11 @@ func (c *Conf) nginxConfFile() string {
 }
 
 // New creates an nginx updater.
-func New(nginxConf Conf) *nginxUpdater {
+func New(nginxConf Conf) controller.Updater {
+	return updaterFromConf(nginxConf)
+}
+
+func updaterFromConf(nginxConf Conf) *nginxUpdater {
 	initMetrics()
 
 	nginxConf.WorkingDir = strings.TrimSuffix(nginxConf.WorkingDir, "/")
--- a/nginx/nginx.go
+++ b/nginx/nginx.go
@@ -13,6 +13,7 @@ import (
 
 	"time"
 
+	"sync"
 	"syscall"
 
 	"errors"
@@ -88,10 +89,15 @@ type nginxUpdater struct {
 	running              util.SafeBool
 	lastErr              util.SafeError
 	metricsUnhealthy     util.SafeBool
-	initialUpdateApplied util.SafeBool
+	initialUpdateApplied initialUpdateApplied
 	doneCh               chan struct{}
 	nginx                *nginx
 	updateRequired       util.SafeBool
+}
+
+type initialUpdateApplied struct {
+	sync.Mutex
+	done bool
 }
 
 // Used for generating nginx config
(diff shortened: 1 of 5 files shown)


nginx/nginx.go
```diff
@@ -191,6 +182,35 @@ func (n *nginxUpdater) initialiseNginxConf() error {
...
 }
 
+func (n *nginxUpdater) ensureNginxRunning() error {
+	// Guard against the (hopefully unlikely) event that two updates could be
+	// received concurrently and attempt to start Nginx twice.
+	n.initialUpdateApplied.Lock()
+	defer n.initialUpdateApplied.Unlock()
+
+	if !n.initialUpdateApplied.done {
+		log.Info("Starting nginx with initial configuration")
```
jsravn: Can we change the comment to "Starting nginx for the first time"? As we no longer have an "initial configuration".

nginx/nginx.go
```diff
@@ -488,9 +509,6 @@ func (n *nginxUpdater) Health() error {
 	if !n.running.Get() {
 		return errors.New("nginx is not running")
 	}
-	if !n.initialUpdateApplied.Get() {
-		return errors.New("waiting for initial update")
-	}
```
jsravn: Can we keep this in, it makes it clearer the state of nginx.

alangibson01:
Changes made as requested.

nginx/nginx.go
alangibson01: Makes sense, that way we know whether Nginx wasn't yet started or whether it died.

nginx/nginx.go
alangibson01: Done.

nginx/nginx_test.go
```diff
@@ -80,10 +93,32 @@ func TestStopWaitsForGracefulShutdownOfNginx(t *testing.T) {
...
 	assert.Error(lb.Health(), "should have waited for nginx to gracefully stop")
 }
 
+func TestNginxNotStartedUntilFirstUpdate(t *testing.T) {
+	tmpDir := setupWorkDir(t)
+	lb := newUpdaterWithBinary(tmpDir, "./fake_graceful_nginx.py")
+
+	assert.NoError(t, lb.Start())
+	assert.Nil(t, lb.nginx.Process)
+}
```
jsravn: Would it be possible to test this without having to access the private state?

jsravn: For instance, we could use `fake_failing_nginx.sh` and make sure it only fails when the first update happens.

alangibson01:
@jsravn Hmmm, that's actually shown up a bug that I'd missed.

alangibson01:
Actually, scratch that - I was confusing the `Update()` function with the `update()` function, which is maybe understandable. I'll rename `update()` to be a bit more specific to what it does.

alangibson01:
Just waiting for the Travis build to go through ... it's queued at the moment.

nginx/nginx.go
```diff
@@ -208,8 +206,6 @@ func (n *nginxUpdater) ensureNginxRunning() error {
 
 		go n.periodicallyUpdateMetrics()
 		go n.backgroundSignaller()
-
-		n.initialUpdateApplied.done = true
```
jsravn: How come this was moved? If we move it, it won't try to restart nginx if it fails due to bad config (maybe that's okay, though - kubelet will kill it eventually).

nginx/nginx_test.go
```diff
@@ -100,23 +100,15 @@ func TestStopWaitsForGracefulShutdownOfNginx(t *testing.T) {
...
-}
-
 func TestNginxStartedAfterFirstUpdate(t *testing.T) {
 	tmpDir := setupWorkDir(t)
-	lb := newUpdaterWithBinary(tmpDir, "./fake_graceful_nginx.py")
+	lb := newUpdaterWithBinary(tmpDir, "./fake_failing_nginx.sh")
 
-	assert.NoError(t, lb.Start())
-	assert.NoError(t, lb.Update(controller.IngressUpdate{Entries: []controller.IngressEntry{{
+	lb.Start()
```
jsravn: Shouldn't we assert NoError here?

nginx/nginx_test.go
```diff
@@ -165,7 +157,10 @@ func TestFailsIfNginxDiesEarly(t *testing.T) {
 	lb := newUpdaterWithBinary(tmpDir, "./fake_failing_nginx.sh")
 
 	assert.Error(lb.Start())
```
jsravn: I wonder why this generates an error. Shouldn't `Start` succeed, since it doesn't start nginx yet?

nginx/nginx_test.go
```diff
@@ -100,23 +100,15 @@ func TestStopWaitsForGracefulShutdownOfNginx(t *testing.T) {
...
 
-	assert.NoError(t, lb.Start())
-	assert.NoError(t, lb.Update(controller.IngressUpdate{Entries: []controller.IngressEntry{{
+	lb.Start()
+	err := lb.Update(controller.IngressUpdate{Entries: []controller.IngressEntry{{
 		Host: "james.com",
-	}}}))
-	assert.NotNil(t, lb.nginx.Process.Pid)
+	}}})
+	assert.EqualError(t, err, "nginx died shortly after starting")
```
jsravn: This test is duplicating  `TestFailsIfNginxDiesEarly`. Instead we could have it execute a non-existent command, that would cause it to fail immediately.

nginx/nginx.go
alangibson01: With it in its previous position, it wasn't possible for the healthcheck to tell the difference between waiting for an initial update to be applied, and Nginx having stopped unexpectedly.

That's maybe a lesser concern, however, and we can find some other way of conveying that information.

nginx/nginx_test.go
alangibson01: It still runs the version check, which returns an error if it fails.

nginx/nginx_test.go
alangibson01: We certainly can do, but I thought it's irrelevant to what the test is supposed to be testing, so better to leave it out.

nginx/nginx_test.go
alangibson01: Indeed, and I had missed how similar these two tests are.

I think the problem is more that we're testing the Nginx startup happens by the side-effect of it failing, and then testing that we receive an error when the launched process exits.

alangibson01:
@jsravn So I've been thinking about a better way to detect that Nginx (well, our "fake" Nginx) is started up only after an update, rather than just expecting something to fail.

I thought it might be an idea to change `fake_graceful_nginx.py` so that it writes out a marker file to the temp directory, and then we can check that's actually been written. Still feels a bit convoluted, but I think it would make for a more understandable test?

What do you think?

nginx/nginx_test.go
jsravn: Ah right. We should add a comment here for that, since it's not obvious. Ideally the script would handle `-v` but fail when simulating a start.

alangibson01: Easy enough to do the -v handling, I'll do that.

alangibson01:
Made the changes discussed above.

nginx/nginx.go
```diff
@@ -509,7 +511,7 @@ func (n *nginxUpdater) Health() error {
 	n.initialUpdateApplied.Lock()
 	defer n.initialUpdateApplied.Unlock()
```
jsravn: We don't need this lock anymore.

nginx/nginx_test.go
```diff
@@ -119,9 +132,13 @@ func TestUnhealthyIfHealthPortIsNotUp(t *testing.T) {
 	lb := newUpdater(tmpDir)
 
 	assert.NoError(lb.Start())
+	assert.NoError(lb.Update(controller.IngressUpdate{Entries: []controller.IngressEntry{{
+		Host: "james.com",
+	}}}))
 
 	time.Sleep(smallWaitTime)
-	assert.Error(lb.Health(), "should be unhealthy")
+
```
jsravn: remove this whitespace?

nginx/nginx.go
```diff
@@ -86,13 +86,14 @@ func (n *nginxUpdater) signalIfRequired() {
...
+	lastErr              	util.SafeError
+	metricsUnhealthy     	util.SafeBool
+	initialUpdateApplied 	initialUpdateApplied
+	initialUpdateAttempted	util.SafeBool
+	doneCh               	chan struct{}
+	nginx                	*nginx
+	updateRequired       	util.SafeBool
 }
 
 type initialUpdateApplied struct {
```
jsravn: I think we could rename this if we have 'initialUpdateAttempted' now. We can call it something like `nginxStarted`.

nginx/nginx.go
```diff
@@ -485,12 +508,15 @@ func createNginxPath(rawPath string) string {
 }
 
 func (n *nginxUpdater) Health() error {
+	n.initialUpdateApplied.Lock()
+	defer n.initialUpdateApplied.Unlock()
```
jsravn: I don't think we need this lock anymore.

nginx/nginx.go
```diff
@@ -85,13 +86,19 @@ func (n *nginxUpdater) signalIfRequired() {
...
+	lastErr                util.SafeError
+	metricsUnhealthy       util.SafeBool
+	initialUpdateApplied   initialUpdateApplied
+	initialUpdateAttempted util.SafeBool
+	doneCh                 chan struct{}
+	nginx                  *nginx
+	updateRequired         util.SafeBool
+}
+
+type initialUpdateApplied struct {
```
jsravn: Can we rename this to `nginxStarted` or something similar. With `initialUpdateAttempted` we can be more specific with this one.

nginx/nginx.go
alangibson01: Yep, you're right.

nginx/nginx.go
alangibson01: I'll go for nginxStarted.

nginx/nginx_test.go
alangibson01: Fair enough.

alangibson01:
Right, hopefully that's that, then.

jsravn:
@alangibson01 Okay over to you to squash it :)

jsravn:
Don't forget to release after merging - can do that by creating the tag and pushing it.

alangibson01:
I would if I had permission to merge :)

And when you say releasing is done by creating a tag and pushing it, I presume you mean just a regular git tag?

PR #144, opened 2017-07-05
Do not delete unassociated A records

lheinlen:
We found ourselves wanting to use `feed-dns` with a Route 53 zone that was already populated with existing records.  To facilitate this, we added a switch to allow `feed-dns` to only clean up records which alias existing load balancers that match feed's selection criteria.  While this does mean the cleanup could potentially leave records pointing to load balancers that have been deleted, we felt that this was a rare enough case to be a fair trade off for the ability to use the service with an existing zone.

### DIFF
--- a/dns/dns_updater.go
+++ b/dns/dns_updater.go
@@ -33,6 +33,7 @@ type updater struct {
 	elbLabelValue string
 	domain        string
 	findELBs      findELBsFunc
+	onlyDelAssoc  bool
 }
 
 // ALB represents the subset of AWS operations needed for dns_updater.go
@@ -41,7 +42,7 @@ type ALB interface {
 }
 
 // New creates an updater for dns
-func New(hostedZoneID, region string, elbLabelValue string, albNames []string, retries int) controller.Updater {
+func New(hostedZoneID, region string, elbLabelValue string, albNames []string, retries int, onlyDelAssoc bool) controller.Updater {
 	initMetrics()
 	session := session.New(&aws.Config{Region: &region})
 	return &updater{
@@ -52,6 +53,7 @@ func New(hostedZoneID, region string, elbLabelValue string, albNames []string, r
 		albNames:      albNames,
 		elbLabelValue: elbLabelValue,
 		findELBs:      elb.FindFrontEndElbs,
+		onlyDelAssoc:  onlyDelAssoc,
 	}
 }
 
@@ -146,6 +148,10 @@ func (u *updater) Update(update controller.IngressUpdate) error {
 		failedCount.Inc()
 		return err
 	}
+
+	if u.onlyDelAssoc {
+		aRecords = u.filterUnassociated(aRecords)
+	}
 	recordsGauge.Set(float64(len(aRecords)))
 
 	changes := u.calculateChanges(aRecords, update)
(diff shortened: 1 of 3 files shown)


dns/dns_updater.go
```diff
@@ -161,6 +167,24 @@ func (u *updater) Update(update controller.IngressUpdate) error {
...
+	for _, dns := range u.schemeToDNS {
+		managedLBs[dns.dnsName] = true
+	}
+	filtered := make([]*route53.ResourceRecordSet, 0)
+	for _, rec := range aRecords {
+		if rec.AliasTarget != nil && rec.AliasTarget.DNSName != nil && managedLBs[*rec.AliasTarget.DNSName] {
+			log.Debugf("Keeping managed A record %s", *rec.Name)
+			filtered = append(filtered, rec)
+		} else {
+			log.Infof("Filtering non-managed A record %s", *rec.Name)
```
jsravn: Can we combine this into a single log line for all filtered records, so it's less spammy.

jsravn: Also it's not necessarily just A records. So let's change the log message to something like `Filtered non-managed resource record set`.

dns/dns_updater.go
```diff
@@ -161,6 +167,24 @@ func (u *updater) Update(update controller.IngressUpdate) error {
...
 
+func (u *updater) filterUnassociated(aRecords []*route53.ResourceRecordSet) []*route53.ResourceRecordSet {
+	managedLBs := make(map[string]bool)
+	for _, dns := range u.schemeToDNS {
+		managedLBs[dns.dnsName] = true
+	}
+	filtered := make([]*route53.ResourceRecordSet, 0)
+	for _, rec := range aRecords {
+		if rec.AliasTarget != nil && rec.AliasTarget.DNSName != nil && managedLBs[*rec.AliasTarget.DNSName] {
+			log.Debugf("Keeping managed A record %s", *rec.Name)
```
jsravn: `Keeping managed resource record set`

jsravn:
@lheinlen Thanks for this, it's a great improvement. Can we remove the flag and just make this the default behaviour? I can't imagine why anyone would want feed-dns to delete the non managed RRS. We can bump the major version on release in case anyone is depending on the behaviour.

dns/dns_updater.go
lheinlen: I'm happy to make these changes.  I would like to ensure that I understand something first to make sure I haven't made some sort of logical error that I'm missing.  This method is filtering the results of `aRecords, err := u.r53.GetARecords()` which should only return A records (as long as my reading of the code is accurate) which is why the messages are as they are.  Are you pointing out that within the context of the `filterUnassociated` method there's nothing that guarantees that they are actually A records or have I missed something larger?

lheinlen:
I will try to get the flag removed and the PR updated today.

dns/dns_updater.go
```diff
@@ -97,7 +97,7 @@ func (u *updater) initELBs() error {
 
 	u.schemeToDNS = make(map[string]dnsDetails)
 	for scheme, lbDetails := range elbs {
-		u.schemeToDNS[scheme] = dnsDetails{dnsName: lbDetails.DNSName, hostedZoneID: lbDetails.HostedZoneID}
+		u.schemeToDNS[scheme] = dnsDetails{dnsName: lbDetails.DNSName + ".", hostedZoneID: lbDetails.HostedZoneID}
```
lheinlen: I noticed that the ELBs in the `schemeToDNS` map did not append a `.` to the DNS Name while the ALBs did.  By making this consistent, the logic in what is now `determineManagedRecordSets` (which was actually not working properly for ELBs in my previous version) can remain a simple hash lookup.  I read through the code and I didn't see anywhere that this would cause an issue but if I have missed something please let me know and I will revert this and take care of it in  `determineManagedRecordSets`.

jsravn: Good catch. That looks like a bug to me, the ELB map should have a trailing dot as well. Otherwise it will always create an UPSERT for every single domain every time (safe, but inefficient). Can you add a
 check that `lbDetails.DNSName` doesn't have a trailing dot first, though? If it does (it shouldn't though), return an error.

jsravn: Can you add a test for it as well? Even if it just calls `initELBs` directly.

lheinlen: Sure, I'll try to get to it this weekend but it might not be until Monday.

jsravn:
Released as v1.1.0

PR #145, opened 2017-07-06
Update README.md with latest versions

jsravn:


alangibson01:
@jsravn Shouldn't we update the version numbers in `examples/feed-dns-deployment.yml` and `examples/feed-ingress-deployment.yml` as well?

jsravn:
@alangibson01 Hmm yeah we could. But this is going to be painful to update all the time. Maybe we should just use latest.

alangibson01:
@jsravn If we can do that, it's probably a good idea. Version history is still going to be recorded in the changelog, if someone wants to see it for reference.

jsravn:
@alangibson01 I updated it everywhere. If it becomes too painful, let's switch to using latest in the examples, and a concrete version in the README.md.

PR #146, opened 2017-07-07
Make deduping ingress entries deterministic.

balooo:
The previous approach tried to order ingress by CreationTimestamp before
picking the most recent ingress.  This did not work because multiple
duplicate ingresses could be created at the same time.

This fix orders ingress entries by Namespace,Name,Host,Path and only
uses the first ingress 'Host/Path' encountered to dedupe.  Kuberentes
guarentees unique ingress for a given 'Namespace/Name' which will make
this deduping deterministic.

fixes: https://github.com/sky-uk/umc-core/issues/3793

### DIFF
--- a/controller/ingress_entry.go
+++ b/controller/ingress_entry.go
@@ -6,9 +6,24 @@ import (
 	"time"
 )
 
-// IngressUpdate data
-type IngressUpdate struct {
-	Entries []IngressEntry
+// IngressEntries type
+type IngressEntries []IngressEntry
+
+// allows us to sort ingress entries.
+func (entries IngressEntries) Len() int {
+	return len(entries)
+}
+
+// allows us to sort ingress entries.
+func (entries IngressEntries) Less(i, j int) bool {
+	iEntry := fmt.Sprintf("%s:%s:%s:%s", entries[i].Namespace, entries[i].Name, entries[i].Host, entries[i].Path)
+	jEntry := fmt.Sprintf("%s:%s:%s:%s", entries[j].Namespace, entries[j].Name, entries[j].Host, entries[j].Path)
+	return iEntry < jEntry
+}
+
+// allows us to sort ingress entries.
+func (entries IngressEntries) Swap(i, j int) {
+	entries[i], entries[j] = entries[j], entries[i]
 }
 
 // IngressEntry describes the ingress for a single host, path, and service.
--- a/controller/ingress_entry.go
+++ b/controller/ingress_entry.go
@@ -9,23 +9,6 @@ import (
 // IngressEntries type
 type IngressEntries []IngressEntry
 
-// allows us to sort ingress entries.
-func (entries IngressEntries) Len() int {
-	return len(entries)
-}
-
-// allows us to sort ingress entries.
-func (entries IngressEntries) Less(i, j int) bool {
-	iEntry := fmt.Sprintf("%s:%s:%s:%s", entries[i].Namespace, entries[i].Name, entries[i].Host, entries[i].Path)
-	jEntry := fmt.Sprintf("%s:%s:%s:%s", entries[j].Namespace, entries[j].Name, entries[j].Host, entries[j].Path)
-	return iEntry < jEntry
-}
-
-// allows us to sort ingress entries.
-func (entries IngressEntries) Swap(i, j int) {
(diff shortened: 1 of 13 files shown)


controller/ingress_entry.go
```diff
@@ -6,9 +6,24 @@ import (
...
+
+// allows us to sort ingress entries.
+func (entries IngressEntries) Len() int {
+	return len(entries)
+}
+
+// allows us to sort ingress entries.
+func (entries IngressEntries) Less(i, j int) bool {
+	iEntry := fmt.Sprintf("%s:%s:%s:%s", entries[i].Namespace, entries[i].Name, entries[i].Host, entries[i].Path)
+	jEntry := fmt.Sprintf("%s:%s:%s:%s", entries[j].Namespace, entries[j].Name, entries[j].Host, entries[j].Path)
```
jsravn: Use `strings.Join`. Also I'm not sure how safe it is to use ":" as the separator. It would be faster/more correct to compare each item separately, but a lot more code.

jsravn: Also this should probably be in nginx.go since it's specific to nginx.

nginx/nginx.go
```diff
@@ -465,7 +465,9 @@ type ingressKey struct {
 	Host, Path string
 }
 
-func uniqueIngressEntries(ingressEntries []controller.IngressEntry) []controller.IngressEntry {
+func uniqueIngressEntries(ingressEntries controller.IngressEntries) []controller.IngressEntry {
+	sort.Sort(ingressEntries)
```
jsravn: Use  `sort.Slice` so you only need the Less function.

controller/ingress_entry.go
balooo: I put this here because it's where IngressUpdate was previously defined, and it's also where the IngressEntry struct is being defined.

balooo: Why do you think in ':' makes it not safe?  I think I actually don't even really need a delimiter actually.  And I don't think the number of ingress we are expecting even in the very very long term would make this a performance problem.

jsravn: It's specific to nginx sorting behaviour - it doesn't make sense to put it in the top level controller package. It should be close to its usage.

balooo: I moved the slice function to nginx.go

nginx/nginx.go
```diff
@@ -465,11 +465,17 @@ type ingressKey struct {
...
 }
 
-func uniqueIngressEntries(ingressEntries controller.IngressEntries) []controller.IngressEntry {
-	sort.Sort(ingressEntries)
+func uniqueIngressEntries(entries controller.IngressEntries) []controller.IngressEntry {
+	sort.Slice(entries, func(i, j int) bool {
+		iEntry := entries[i]
+		jEntry := entries[j]
+		iString := strings.Join([]string{iEntry.Namespace, iEntry.Name, iEntry.Host, iEntry.Path}, "")
+		jString := strings.Join([]string{jEntry.Namespace, jEntry.Name, jEntry.Host, jEntry.Path}, "")
```
jsravn: You still need a separator here, something that won't be used in any of the strings. Otherwise "foo/bar" and "foob/ar" are equivalent.

nginx/nginx.go
```diff
@@ -465,36 +465,32 @@ type ingressKey struct {
 	Host, Path string
 }
 
-func uniqueIngressEntries(ingressEntries []controller.IngressEntry) []controller.IngressEntry {
+func uniqueIngressEntries(entries controller.IngressEntries) []controller.IngressEntry {
+	sort.Slice(entries, func(i, j int) bool {
+		iEntry := entries[i]
+		jEntry := entries[j]
+		iString := strings.Join([]string{iEntry.Namespace, iEntry.Name, iEntry.Host, iEntry.Path}, ":^^:")
+		jString := strings.Join([]string{jEntry.Namespace, jEntry.Name, jEntry.Host, jEntry.Path}, ":^^:")
```
jsravn: I think a single colon is good here :).

balooo: You just said you didn't want that :smile:

totahuanocotl:
@balooo ready for you

totahuanocotl:
@baloo I think this was missing the changelog. Can you add it in a new PR?

PR #147, opened 2017-07-11
Update CHANGELOG.md - release 2.0.0

jsravn:


jsravn:
We discussed that 1.1.0 makes more sense, as this shouldn't break prior behaviour (no one should be relying on it deleting all non managed records).

PR #151, opened 2017-08-16
Avoid duplicate deregistration delay in both ALB/ELB

totahuanocotl:
Avoid waiting deregistration time for alb/elb if there were no attached
instances, even if one was provided.

### DIFF
--- a/alb/alb.go
+++ b/alb/alb.go
@@ -103,7 +103,10 @@ func (a *alb) Stop() error {
 		}
 	}
 
-	time.Sleep(a.targetGroupDeregistrationDelay)
+	if len(a.albARNs) > 0 {
+		log.Infof("Waiting %vs to finish ALB target group deregistration", a.targetGroupDeregistrationDelay)
+		time.Sleep(a.targetGroupDeregistrationDelay)
+	}
 
 	return nil
 }
--- a/alb/alb.go
+++ b/alb/alb.go
@@ -20,7 +20,10 @@ import (
 )
 
 // New creates a controller.Updater for attaching to ALB target groups on first update.
-func New(region string, targetGroupNames []string, targetGroupDeregistrationDelay time.Duration) controller.Updater {
+func New(region string, targetGroupNames []string, targetGroupDeregistrationDelay time.Duration) (controller.Updater, error) {
+	if len(targetGroupNames) == 0 {
+		return nil, fmt.Errorf("Unable to create Alb Updater: missing target group names")
+	}
 	initMetrics()
 	log.Infof("ALB frontend region: %s target groups: %v", region, targetGroupNames)
 	session := session.New(&aws.Config{Region: &region})
@@ -31,7 +34,7 @@ func New(region string, targetGroupNames []string, targetGroupDeregistrationDela
 		targetGroupDeregistrationDelay: targetGroupDeregistrationDelay,
 		region:      region,
 		initialised: initialised{},
-	}
+	}, nil
 }
 
 type alb struct {
@@ -103,10 +106,8 @@ func (a *alb) Stop() error {
 		}
 	}
 
-	if len(a.albARNs) > 0 {
-		log.Infof("Waiting %vs to finish ALB target group deregistration", a.targetGroupDeregistrationDelay)
-		time.Sleep(a.targetGroupDeregistrationDelay)
-	}
+	log.Infof("Waiting %vs to finish ALB target group deregistration", a.targetGroupDeregistrationDelay)
+	time.Sleep(a.targetGroupDeregistrationDelay)
 
 	return nil
 }
--- a/alb/alb.go
+++ b/alb/alb.go
@@ -4,6 +4,7 @@ Package alb provides an updater for an ALB frontend to attach nginx to.
 package alb
 
 import (
+	"errors"
 	"fmt"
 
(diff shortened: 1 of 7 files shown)


jsravn:
@totahuanocotl Would it be better to just not create an ALB updater if ALB targetGroupNames is empty? We should make it an invariant in alb.New that targetGroupNames is non-empty as well. Same with the ELB side.

totahuanocotl:
@jsravn I think we cannot manage both elb/alb the same way. For ELB the default value for the frontEnds to attach to is "", which means, don't attach. For ALB, there is no default value, and the absence means, don't attach. 
For ALBs we can do that check in alb.New, but for ELB we can't fail if it is empty. and we can't fail if the expected elb number is 0 either.
I tried looking for a way to create an optional flag in go, but it seems it is not possible. I could go for one of the following options:
1) Use additional flags to turn alb/elb on and  use those flags AND the target frontend values (defaulted to nothing) to do the check. Fail if they are empty.  
2) If no values are provided create a Noop updater for elb/alb. 
3) Treat elb and alb differently, and for ALB fail on New, and for ELB leave it as it is.

jsravn:
@totahuanocotl Not sure I follow. Why can't you just check in main.go if frontends is empty, then don't create the elb instance at all?

totahuanocotl:
@jsravn applied the changes

alb/alb.go
```diff
@@ -103,6 +106,7 @@ func (a *alb) Stop() error {
 		}
 	}
 
+	log.Infof("Waiting %vs to finish ALB target group deregistration", a.targetGroupDeregistrationDelay)
```
jsravn: Remove the `s`, duration already displays units.

alb/alb.go
```diff
@@ -20,7 +20,10 @@ import (
 )
 
 // New creates a controller.Updater for attaching to ALB target groups on first update.
-func New(region string, targetGroupNames []string, targetGroupDeregistrationDelay time.Duration) controller.Updater {
+func New(region string, targetGroupNames []string, targetGroupDeregistrationDelay time.Duration) (controller.Updater, error) {
+	if len(targetGroupNames) == 0 {
+		return nil, fmt.Errorf("Unable to create Alb Updater: missing target group names")
```
jsravn: Errors should always be lowercase. Also use `error.New` if you don't need to format a string.

controller/controller.go
```diff
@@ -235,7 +235,7 @@ func (c *controller) Stop() error {
 	close(c.doneCh)
 
 	for i := range c.updaters {
-		u := c.updaters[len(c.updaters)-i-1]
+		u := c.updaters[len(c.updaters)-1-i]
```
jsravn: How come?

elb/elb.go
```diff
@@ -25,7 +25,10 @@ import (
 const ElbTag = "sky.uk/KubernetesClusterFrontend"
 
 // New creates a new ELB frontend
-func New(region string, labelValue string, expectedNumber int, drainDelay time.Duration) controller.Updater {
+func New(region string, labelValue string, expectedNumber int, drainDelay time.Duration) (controller.Updater, error) {
+	if labelValue == "" {
+		return nil, fmt.Errorf("Unable to create Elb Updater: Missing label value for the tag %v", ElbTag)
```
jsravn: Same as prior comment on errors.

elb/elb.go
```diff
@@ -221,6 +220,7 @@ func (e *elb) Stop() error {
 		return errors.New("at least one ELB failed to detach")
 	}
 
+	log.Infof("Waiting %vs to finish ELB deregistration", e.drainDelay)
```
jsravn: Same as prior comment.

alb/alb_test.go
```diff
@@ -108,27 +108,20 @@ func (m *mockMetadata) mockInstanceMetadata(instanceID string) {
...
 	//when
-	a.Start()
-	a.Update(controller.IngressEntries{})
-	a.Stop()
+	_, err := New(region, []string{}, time.Nanosecond)
 
 	//then
-	mockElb.AssertExpectations(t)
-	mockMetadata.AssertExpectations(t)
+	assert.EqualError(t, err, "Unable to create Alb Updater: missing target group names")
```
jsravn: I don't think we should assert against the exact error message, just assert the error is not nil.

elb/elb_test.go
```diff
@@ -130,27 +130,20 @@ func mockInstanceMetadata(mockMd *fakeMetadata, instanceID string) {
...
 	//when
-	e.Start()
-	e.Update(controller.IngressEntries{})
-	e.Stop()
+	_, err := New(region, "", 1, 0)
 
 	//then
-	mockElb.AssertExpectations(t)
-	mockMetadata.AssertExpectations(t)
+	assert.EqualError(t, err, "Unable to create Elb Updater: Missing label value for the tag "+ElbTag)
```
jsravn: I don't think we should assert against the exact error message.`

jsravn:
@totahuanocotl Also need a CHANGELOG.md update.

controller/controller.go
totahuanocotl: ah, i had made a refactor `of last := len(c.updaters)-1` but then i undid it

PR #153, opened 2017-08-23
[WIP] support commandline cnames when you don't want ELBs

shin-nien:


cmd/feed-dns/main.go
```diff
@@ -67,6 +69,10 @@ func init() {
 		"A label=value pair to attach to metrics pushed to prometheus. Specify multiple times for multiple labels.")
 	flag.IntVar(&awsAPIRetries, "aws-api-retries", defaultAwsAPIRetries,
 		"Number of times a request to the AWS API is retried.")
+	flag.StringVar(&internalIngressFQDN, "internal-ip", "",
```
lw346: These are expected to be CNAMEs not IPs, so the variable name and help will need updating to reflect this

cmd/feed-dns/main.go
```diff
@@ -104,4 +116,9 @@ func validateConfig() {
 		log.Error("Must supply r53-hosted-zone")
 		os.Exit(-1)
 	}
+
+	if (internalIngressFQDN != "") != (externalIngressFQDN != "") { //XOR
```
lw346: While it may be correct, this is very confusing syntax and could lead to maintainability issues in future - you may disagree, but I would recommend making this more explicit:

`if (internalIngressFQDN == "" && externalIngressFQDN != "") || (internalIngressFQDN != "" && externalIngressFQDN == "") {`

lw346: This check should also be before we create the controller and dns updater.

lw346: We should probably also have a check for argument exclusivity - given that if these parameters are present we will ignore the elb label argument, we should fail to run if both are specified.

dns/dns_updater.go
```diff
@@ -144,17 +159,20 @@ func (u *updater) Health() error {
 }
 
 func (u *updater) Update(entries controller.IngressEntries) error {
-	aRecords, err := u.r53.GetARecords()
+	route53Records, err := u.r53.GetRecords()
 	if err != nil {
 		log.Warn("Unable to get A records from Route53. Not updating Route53.", err)
```
lw346: Needs updating to be not A-record specific

dns/dns_updater.go
```diff
@@ -144,17 +159,20 @@ func (u *updater) Health() error {
...
 		failedCount.Inc()
 		return err
 	}
 
-	aRecords = u.determineManagedRecordSets(aRecords)
-	recordsGauge.Set(float64(len(aRecords)))
+	// Flatten Alias (A) and CNAME records into a common structure
+	var records []record = u.flattenRoute53ResourceRecordSet(route53Records)
```
lw346: `records := u.flattenRoute53ResourceRecordSet(route53Records)` would be the canonical syntax here I think?

dns/dns_updater.go
```diff
@@ -167,18 +185,39 @@ func (u *updater) Update(entries controller.IngressEntries) error {
 	return nil
 }
 
-func (u *updater) determineManagedRecordSets(rrs []*route53.ResourceRecordSet) []*route53.ResourceRecordSet {
+func (u *updater) flattenRoute53ResourceRecordSet(rrs []*route53.ResourceRecordSet) []record {
+	records := make([]record, 0)
```
lw346: I think making a 0-length array of records is quite confusing here.

You can instead declare `var records []record`, and the `append` function will `make()` if it needs to

cmd/feed-dns/main.go
shin-nien: Done

cmd/feed-dns/main.go
shin-nien: Agreed. Done.

dns/dns_updater.go
shin-nien: Done. lint also failed this.

dns/dns_updater.go
shin-nien: Agreed. Changed other instances of this.

cmd/feed-dns/main.go
```diff
@@ -104,4 +116,14 @@ func validateConfig() {
...
 		os.Exit(-1)
 	}
+
+	if (internalCNAME != "" || externalCNAME != "") && elbLabelValue != "" {
+		log.Error("Can't supply both ELB label and non-ELB CNAME. Choose one or the other.")
+		os.Exit(-1)
+	}
+
+	if (internalCNAME == "" && externalCNAME != "") || (internalCNAME != "" && externalCNAME == "") {
+		log.Error("Must supply both internal-ip and external-ip if any are to be provided.")
```
lw346: Forgot to update the argument names here :)

cmd/feed-dns/main.go
```diff
@@ -104,4 +116,14 @@ func validateConfig() {
 		log.Error("Must supply r53-hosted-zone")
 		os.Exit(-1)
 	}
+
+	if (internalCNAME != "" || externalCNAME != "") && elbLabelValue != "" {
```
lw346: Probably want checks on command line arguments before you start using them, so these if blocks should move to the top of `main()`

controller/controller.go
```diff
@@ -145,19 +146,27 @@ func (c *controller) updateIngresses() error {
...
-						Allow:                   c.defaultAllow,
-						ELbScheme:               ingress.Annotations[frontendElbSchemeAnnotation],
+						Namespace:      ingress.Namespace,
+						Name:           ingress.Name,
+						Host:           rule.Host,
+						Path:           path.Path,
+						ServiceAddress: address,
+						ServicePort:    int32(path.Backend.ServicePort.IntValue()),
+						Allow:          c.defaultAllow,
+						//ELbScheme:               ingress.Annotations[frontendElbSchemeAnnotation],
```
lw346: Should be removed rather than commented out

controller/controller.go
```diff
@@ -145,19 +146,27 @@ func (c *controller) updateIngresses() error {
...
 						StripPaths:              c.defaultStripPath,
 						BackendKeepAliveSeconds: c.defaultBackendTimeout,
 						CreationTimestamp:       ingress.CreationTimestamp.Time,
 					}
 
+					log.Infof("ingress: %s", ingress)
```
lw346: There is a distinct lack of logging in this project, so 👍 for adding more, but it would be nice to know *why* this is being logged in the log message - ie. "discovered ingress: %s" or "ingress to add DNS for: %s"

cmd/feed-dns/main.go
lw346: I suck, and can't read context - this already is at the top.

cmd/feed-dns/main.go
shin-nien: Done.

cmd/feed-dns/main.go
shin-nien: validate() already gets called at the top of main() (or near enough)

controller/controller.go
shin-nien: Done

shin-nien:
related to https://github.com/sky-uk/umc-core/issues/4062

jsravn:
CHANGELOG.md needs updating too.

README.md
```diff
@@ -59,17 +59,19 @@ See the command line options with:
...
 ### DNS records
 
-The feed-dns controller assumes that it can overwrite any entry in the supplied DNS zone and manages ALIAS
+The feed-dns controller assumes that it can overwrite any entry in the supplied DNS zone and manages ALIAS and CNAME
 records per ingress.
 
 On startup, all ingress entries are queried and compared to all the Record Sets in the configured hosted zone.
 
-Any ALIAS records pointing to one of the ELBs associated with this controller that do not have an ingress
-entry are deleted. For any new ingress entry, an ALIAS record is created to point to the correct ELB. Existing
```
jsravn: I don't think we should delete this part, as these details are pretty important (and just got added in a recent PR). We should mention the "cname" behaviour separately I think.

README.md
```diff
@@ -59,17 +59,19 @@ See the command line options with:
...
+Any records pointing to one of the endpoints associated with this controller that do not have an ingress
+entry are deleted. For any new ingress entry, a record is created to point to the correct endpoint. Existing
 records which do not meet these conditions remain untouched.
 
-Each ingress must have the following tag `sky.uk/frontend-elb-scheme` set to `internal` or `internet-facing` so the A
-record can be set to the correct ELB.
+Each ingress must have the following tag `sky.uk/frontend-scheme` (`sky.uk/frontend-elb-scheme` is **deprecated**) set to `internal` or `internet-facing` so the
+record can be set to the correct endpoint.
```
jsravn: I thought internal/external were purely an  ELB concept. Do we need to extend it generically? Can't we just point to a single IP for the cname case?

README.md
```diff
@@ -50,7 +50,7 @@ Another is to place an SSL termination EC2 instance in front of the ELBs.
 
 ## feed-dns
 
-`feed-dns` manages a Route53 hosted zone, updating entries to point to the correct ELBs. It is designed to be run as a
+`feed-dns` manages a Route53 hosted zone, updating entries to point to ELBs or CNAMEs of your choosing. It is designed to be run as a
```
jsravn: Wording is slightly wrong. CNAMEs point to other records, rather than pointing to a CNAME. Should be something like "updating entries to point to ELBs or arbitrary hostnames".

cmd/feed-dns/main.go
```diff
@@ -67,6 +69,10 @@ func init() {
 		"A label=value pair to attach to metrics pushed to prometheus. Specify multiple times for multiple labels.")
 	flag.IntVar(&awsAPIRetries, "aws-api-retries", defaultAwsAPIRetries,
 		"Number of times a request to the AWS API is retried.")
+	flag.StringVar(&internalCNAME, "internal-cname", "",
+		"the CNAME of the internal facing load-balancer. If specified, external-cname must also be given.")
```
jsravn: Wording is wrong again, it's the hostname of the load balancer. A CNAME is just a type of dns record that aliases to another hostname. I would guess we'd want to support IPs here as well.

cmd/feed-dns/main.go
```diff
@@ -67,6 +69,10 @@ func init() {
...
+	flag.StringVar(&externalCNAME, "external-cname", "",
+		"the CNAME of the internet facing load-balancer. If specified, internal-cname must also be given.")
```
jsravn: I think it'd be better if we did this as a mapping, so you could map internet-scheme to hostname in case we wanted more than the two ELB types. Or if we wanted to use different names.

Maybe that's overcomplicating things though, I'm not sure. The annotation sky.uk/frontend-scheme can be anything though, which seems to suggest this should be parameterised as a map.

cmd/feed-dns/main.go
```diff
@@ -81,7 +87,13 @@ func main() {
...
 	}
 
-	dnsUpdater := dns.New(r53HostedZone, elbRegion, elbLabelValue, albNames, awsAPIRetries)
+	var dnsUpdater controller.Updater
+	if internalCNAME != "" && externalCNAME != "" {
+		addressesWithScheme := map[string]string{"internal": internalCNAME, "internet-facing": externalCNAME}
+		dnsUpdater = dns.New(r53HostedZone, elbRegion, addressesWithScheme, "", albNames, awsAPIRetries)
+	} else {
+		dnsUpdater = dns.New(r53HostedZone, elbRegion, nil, elbLabelValue, albNames, awsAPIRetries)
+	}
```
jsravn: Now that we have optional fields, we should probably add a config struct to pass to New.

cmd/feed-dns/main.go
```diff
@@ -104,4 +116,14 @@ func validateConfig() {
...
 	}
+
+	if (internalCNAME != "" || externalCNAME != "") && elbLabelValue != "" {
+		log.Error("Can't supply both ELB label and non-ELB CNAME. Choose one or the other.")
+		os.Exit(-1)
+	}
+
+	if (internalCNAME == "" && externalCNAME != "") || (internalCNAME != "" && externalCNAME == "") {
+		log.Error("Must supply both internal-cname and external-cname if any are to be provided.")
+		os.Exit(-1)
```
jsravn: I think this is too strict. I could imagine someone only wanting to use an external host for example.

jsravn: It might be better to move the validation inside of `dns.New` so we can unit test it.

README.md
shin-nien: My understanding is that the internal is for anyone in corp to freely hit services using abc.bskyb.com/x but the external only accepts traffic from Akamai IPs; where Akamai also maps p.sky.com/abc/x/y/z back to abc.bskyb.com/x.

README.md
shin-nien: I haven't deleted this part completely; I've just swapped out the AWS specific terms for generic ones because the behaviour is the same whether you're talking ELBs or non-ELBs. At the bottom, I try to make it clear that the record types are different depending on whether an ELB is used or not.

But if that's still confusing then I'm happy to go back and add another paragraph for non-ELBs.

dns/dns_updater.go
```diff
@@ -62,17 +75,19 @@ func (u *updater) String() string {
 func (u *updater) Start() error {
 	log.Info("Starting dns updater")
 
-	if u.elbLabelValue != "" && len(u.albNames) > 0 {
-		return fmt.Errorf("can't specify both elb label value (%s) and alb names (%v) - only one or the other may be"+
-			" specified", u.elbLabelValue, u.albNames)
-	}
+	if len(u.schemeToDNS) == 0 {
```
jsravn: I don't think you need this conditional, as everything in the block is a noop since elbLabelValue == "" and len(u.albNames) == 0.

dns/dns_updater.go
```diff
@@ -167,18 +185,39 @@ func (u *updater) Update(entries controller.IngressEntries) error {
...
-func (u *updater) determineManagedRecordSets(rrs []*route53.ResourceRecordSet) []*route53.ResourceRecordSet {
+func (u *updater) flattenRoute53ResourceRecordSet(rrs []*route53.ResourceRecordSet) []record {
+	var records []record
+
+	for _, recordSet := range rrs {
+		if len(recordSet.ResourceRecords) == 1 {
+			records = append(records, record{
+				name:     *recordSet.Name,
+				pointsTo: *recordSet.ResourceRecords[0].Value,
+			})
```
jsravn: How come we special case `len(recordSet.ResourceRecords) == 1`? Won't this break in arbitrary scenarios?

dns/dns_updater.go
```diff
@@ -167,18 +185,39 @@ func (u *updater) Update(entries controller.IngressEntries) error {
...
 	for _, dns := range u.schemeToDNS {
 		managedLBs[dns.dnsName] = true
 	}
-	managed := make([]*route53.ResourceRecordSet, 0)
-	nonManaged := make([]string, 0)
+	var managed []record
+	var nonManaged []string
 	for _, rec := range rrs {
-		if rec.AliasTarget != nil && rec.AliasTarget.DNSName != nil && managedLBs[*rec.AliasTarget.DNSName] {
+		if rec.name != "" && managedLBs[rec.pointsTo] {
```
jsravn: This logic seems wrong to me - won't it will break if anyone has non alias records in their zone? We need to support any arbitrary record. So this would need to check if it matches one of our static hostnames.

README.md
shin-nien: It still applies if you have other internal services that need a more direct route and your external needs to be fronted by a CDN for example.

We could however make them optional i.e. have one or the other (or both), would that make more sense?

cmd/feed-dns/main.go
jsravn: Also - wouldn't we want to support multiple IPs for redundancy?

dns/dns_updater.go
```diff
@@ -167,18 +185,39 @@ func (u *updater) Update(entries controller.IngressEntries) error {
 	return nil
 }
 
-func (u *updater) determineManagedRecordSets(rrs []*route53.ResourceRecordSet) []*route53.ResourceRecordSet {
+func (u *updater) flattenRoute53ResourceRecordSet(rrs []*route53.ResourceRecordSet) []record {
```
jsravn: Is it flattening? Isn't it just converting to our own structure?

dns/dns_updater.go
```diff
@@ -264,31 +303,39 @@ func (u *updater) createChanges(hostToIngress hostToIngress,
...
 			},
-		},
+		}
+	}
+
+	return &route53.Change{
+		Action:            aws.String(action),
+		ResourceRecordSet: set,
 	}
 }
```
jsravn: Since there is a large amount of conditional logic in here now, I wonder if we should split up this file into two implementations. One for the ELB/ALBs, and one for the static hostnames. I think it'd make it easier to understand as well.

README.md
shin-nien: Agreed. Hostname is a better name.

cmd/feed-dns/main.go
shin-nien: We don't have a use case for IPs and they work significantly different to everything else because you can multiple IPs for an A record so we skipped implementing them.

dns/dns_updater.go
shin-nien: Arbitrary record types wasn't within the scope but I think where we flatten/convert the records could be made much clearer to show what records we are supporting (as referenced by https://github.com/sky-uk/feed/pull/153/files/11f4334038e55b10b9358fab4dd81f44a90ee8e7#r135288514)

dns/dns_updater.go
shin-nien: yes, you're right.

It should be explicit and look for the types ALIAS and CNAME

cmd/feed-dns/main.go
shin-nien: The annotation's value can't be anything; in the source it's very much tied to just 'internal' or 'internet-facing'.

cmd/feed-dns/main.go
shin-nien: Agreed. You shouldn't have to specify both, they can be optional.

cmd/feed-dns/main.go
shin-nien: The scheme is limited to `internal` and `internet-facing` (which is an ELB concept like you say above) and although the annotation has been renamed to be more general I don't think there's a need to extend it any further.

shin-nien:
On hold until further notice

PR #158, opened 2017-11-10
Stop logging out the entire Route53 record set on update at Info

lw346:
Bump the entire record set log down to Debug, and add a new Info logging line with the number of current records.


lw346:
I've declared intent to release this as v1.2.2 and updated the changelog.  When merged, I'll push the v1.2.2 tag to release as per the README.

PR #161, opened 2017-11-23
Add all gorb backend healthcheck types

sebbonnet:
Support all the healthcheck types available in gorb: http, tcp and none.

Also remove deprecated elb-drain-delay ingress flag, which was replaced by drain-delay since 1.4.0.
Reverted default `defaultDrainDelay` to 60s as it inadvertently changed when deprecating `elb-drain-delay`.


cmd/feed-ingress/main.go
```diff
@@ -281,6 +276,10 @@ func createIngressUpdaters() ([]controller.Updater, error) {
 			return nil, fmt.Errorf("invalid gorb services definition. Must be a comma separated list - e.g. 'http-proxy:80,https-proxy:443', but was %s", gorbServicesDefinition)
 		}
 
+		if gorbBackendHealthcheckType != "tcp" && gorbBackendHealthcheckType != "http" && gorbBackendHealthcheckType != "none" {
+			return nil, fmt.Errorf("invalid gorb backend healtcheck type. Must be either 'tcp', 'http' or 'none', but was %s", gorbBackendHealthcheckType)
```
lw346: typo in "healthcheck"

sebbonnet: ok

sebbonnet:
@lw346 none of the flags are mentioned in the README.md, but there are 2 examples for feed-ingress including one for gorb that show all the possible flags you can use - this is where I added the new flag.

PR #162, opened 2017-11-27
Add https parameter + ssl termination

rewiko:
Give the ability to feed-ingress to handle ssl traffic.


cmd/feed-ingress/main.go
```diff
@@ -168,6 +174,9 @@ func init() {
 		"Comma separated list of CIDRs to trust when determining the client's real IP from "+
 			"frontends. The client IP is used for allowing or denying ingress access. "+
 			"This will typically be the ELB subnet.")
+	flag.StringVar(&nginxDefaultSslPath, "default-ssl-path", defaultNginxDefaultSslPath,
```
totahuanocotl: shouldn't this arg just be ssl-path, and have a default value, instead of being called default-ssl-path?

README.md
```diff
@@ -43,11 +42,24 @@ See the command line options with:
 
 ### SSL/TLS
 
-`feed-ingress` expects your ELBs to terminate SSL traffic. We believe this is the safest and best performing
+## SSL termination on ELB
+SSL termination could be done on ELBs, probably the safest and best performing
```
lw346: "SSL termination could be done on ELBS, and we believe that this is the safest and best performing"

README.md
```diff
@@ -43,11 +42,24 @@ See the command line options with:
...
 approach for production usage. Unfortunately, ELBs don't support SNI at this time, so this limits SSL usage to
 a single domain. One workaround is to use a wildcard certificate for the entire zone that `feed-dns` manages.
 Another is to place an SSL termination EC2 instance in front of the ELBs.
 
+## SSL termination on feed-ingress
+
+SSL termination could be done on feed-ingress. You will be able handle the secure websocket and HTTPS with a layer 4 in front (ELB, IPVS(GORB)).
```
lw346: "can be done" ... "This approach still requires a layer 4 load balancer, eg. ELB or IPVS with GORB, in front".

README.md
```diff
@@ -43,11 +42,24 @@ See the command line options with:
...
+
+For the moment you can setup a default wildcard ssl:
+```
+ # Set default ssl path + name file without extension - expected default-ssl.crt and default-ssl.key into /etc/ssl/default-ssl/
```
lw346:
```
# Set default ssl path + name file without extension.  Feed expects two files: one ending in .crt (the CA) and the other in .key (the private key), for example:
# -default-ssl-path=/etc/ssl/default-ssl/default-ssl
# ... will cause feed to look for /etc/ssl/default-ssl.crt and /etc/ssl/default.ssl.key
```

cmd/feed-ingress/main.go
```diff
@@ -43,6 +44,7 @@ var (
 	nginxConfig                    nginx.Conf
 	nginxLogHeaders                cmd.CommaSeparatedValues
 	nginxTrustedFrontends          cmd.CommaSeparatedValues
+	nginxDefaultSslPath            string
```
lw346: This is inconsistent capitalisation - you have HTTPS for an acronym above, but Ssl here

nginx/nginx.tmpl
```diff
@@ -4,6 +4,9 @@ daemon off;
 error_log stderr {{ .LogLevel }};
 pid {{ .WorkingDir }}/nginx.pid;
 
+{{- $DefaultSslPath := .DefaultSslPath }}
+{{- $IngressPorts := .IngressPorts }}
```
lw346: These are both used once directly, so is this aliasing necessary?

nginx/nginx.tmpl
```diff
@@ -138,16 +152,30 @@ http {
...
+        ssl_certificate_key  {{ $DefaultSslPath }}.key;
+        # secure settings (A+ at SSL Labs ssltest at time of writing)
+        # see https://wiki.mozilla.org/Security/Server_Side_TLS#Nginx
+        ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
+        ssl_ciphers 'ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA256:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES128-SHA:DHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:DHE-RSA-CAMELLIA256-SHA:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-SEED-SHA:DHE-RSA-CAMELLIA128-SHA:HIGH:!aNULL:!eNULL:!LOW:!3DES:!MD5:!EXP:!PSK:!SRP:!DSS';
+        ssl_prefer_server_ciphers on;
+        ssl_session_cache shared:SSL:60m;
+        ssl_session_timeout 5m;
+{{- end }}
+       location / {
```
lw346: need an extra space here to satisfy my OCD

nginx/nginx.tmpl
rewiko: Apparently inside a loop you lose the access to the variabe begining by "." so we need to set as template variable.

cmd/feed-ingress/main.go
```diff
@@ -250,8 +259,13 @@ func main() {
 }
 
 func createIngressUpdaters() ([]controller.Updater, error) {
-	nginxConfig.IngressPort = ingressPort
+	nginxConfig.IngressPorts = append(nginxConfig.IngressPorts, nginx.IngressPortConf{Name: "http", Port: ingressPort})
```
jsravn: The append is unnecessary here. Just do `nginxConfig.IngressPorts := []IngressPortConf{{Name: "http", Port: port}}`.

docker/ingress/build-nginx.sh
```diff
@@ -18,6 +18,25 @@ apt-get install --no-install-suggests --no-install-recommends -y \
...
+touch hashes_openssl
+echo "${OPENSSL_SHA256} ${OPENSSL_tarball}" >> hashes_openssl
+if ! sha256sum -c hashes_openssl; then
+    echo "sha256 hashes do not match downloaded files"
+    exit 1
+fi
+
+tar xzf ${OPENSSL_tarball}
+mv openssl-OpenSSL_${OPENSSL} openssl
+
```
jsravn: How come we're downloading the tarball? Why not install with apt, so we can make sure we're getting the latest bugfix on ubuntu?

jsravn: We're also going to need some sort of check mechanism to make sure we keep this up to date...

examples/feed-ingress-deployment-ssl.yml
```diff
@@ -0,0 +1,161 @@
...
+        - containerPort: 12082
+          name: health
+          protocol: TCP
+
+        args:
+        # Ingress nginx port that ELBs will direct traffic towards.
+        - -ingress-port=8080
+
+        # Ingress nginx port for ssl traffic
+        - -ingress-https-port=443
```
jsravn: Probably want 8443 here since 443 won't be allowed by default.

nginx/nginx.go
```diff
@@ -30,6 +30,12 @@ const (
 	metricsUpdateInterval = time.Second * 10
 )
 
+// IngressPortConf configuration
+type IngressPortConf struct {
```
jsravn: Avoid smurf naming, just call the struct "PortConf". Or simply "Port", so it will be `nginx.Port` when used elsewhere.

nginx/nginx.tmpl
jsravn: Aliasing should be moved close to where it's used, rather than at the top of the file.

nginx/nginx.tmpl
```diff
@@ -102,9 +104,21 @@ http {
 
 {{- range $entry := .Servers }}
     # ingress: {{ $entry.Name }}
+  {{- range $IngressPortConf := $IngressPorts }}
```
jsravn: just call it `$portConf` rather than `$IngressPortConf`.

nginx/nginx.tmpl
```diff
@@ -102,9 +104,21 @@ http {
...
     server {
-        listen {{ $port }}{{ if $proxyprotocol }} proxy_protocol{{ end }};
+        listen {{ $IngressPortConf.Port }}{{ if $proxyprotocol }} proxy_protocol{{ end }}{{- if eq $IngressPortConf.Name "https" }} ssl {{ end }};
```
jsravn: Does proxy_protocol work with the ssl option?

nginx/nginx.tmpl
```diff
@@ -102,9 +104,21 @@ http {
...
+        ssl_certificate     {{ $DefaultSslPath }}.crt;
+        ssl_certificate_key  {{ $DefaultSslPath }}.key;
+        # secure settings (A+ at SSL Labs ssltest at time of writing)
+        # see https://wiki.mozilla.org/Security/Server_Side_TLS#Nginx
+        ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
+        ssl_ciphers 'ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA256:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES128-SHA:DHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:DHE-RSA-CAMELLIA256-SHA:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-SEED-SHA:DHE-RSA-CAMELLIA128-SHA:HIGH:!aNULL:!eNULL:!LOW:!3DES:!MD5:!EXP:!PSK:!SRP:!DSS';
+        ssl_prefer_server_ciphers on;
+        ssl_session_cache shared:SSL:60m;
+        ssl_session_timeout 5m;
+{{- end }}
```
jsravn: Please put these options at the top level so we don't have to duplicate them down below in "default backend".

nginx/nginx.tmpl
```diff
@@ -102,9 +104,21 @@ http {
...
+  {{- range $IngressPortConf := $IngressPorts }}
     server {
-        listen {{ $port }}{{ if $proxyprotocol }} proxy_protocol{{ end }};
+        listen {{ $IngressPortConf.Port }}{{ if $proxyprotocol }} proxy_protocol{{ end }}{{- if eq $IngressPortConf.Name "https" }} ssl {{ end }};
         server_name {{ $entry.ServerName }};
+{{- if eq $IngressPortConf.Name "https" }}
+        ssl_certificate     {{ $DefaultSslPath }}.crt;
+        ssl_certificate_key  {{ $DefaultSslPath }}.key;
+        # secure settings (A+ at SSL Labs ssltest at time of writing)
+        # see https://wiki.mozilla.org/Security/Server_Side_TLS#Nginx
```
jsravn: Please link the exact profile you used so we can update it in the future easily.

jsravn: https://mozilla.github.io/server-side-tls/ssl-config-generator/ - Nginx, Modern Profile is what we should use. Possibly adding TLSv1, TLSv1.1 as additional SSL protocols.

jsravn: If possible we could eliminate TLSv1 and TLS v1.1. We only really need to support those if we're going to allow direct connections from old client devices. Everything via Akamai or JVM 8 should be fine with TLS v1.2.

nginx/nginx.tmpl
```diff
@@ -102,9 +104,21 @@ http {
...
+        ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
+        ssl_ciphers 'ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA256:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES128-SHA:DHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:DHE-RSA-CAMELLIA256-SHA:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-SEED-SHA:DHE-RSA-CAMELLIA128-SHA:HIGH:!aNULL:!eNULL:!LOW:!3DES:!MD5:!EXP:!PSK:!SRP:!DSS';
+        ssl_prefer_server_ciphers on;
+        ssl_session_cache shared:SSL:60m;
+        ssl_session_timeout 5m;
```
jsravn: These values don't seem to match the mozilla config. We also want     `ssl_session_tickets off;` so we get forward security.

nginx/nginx.tmpl
rewiko: I've tried but nginx complains,  you can't define the ssl args on the top level.

nginx/nginx.tmpl
rewiko: It sould https://www.nginx.com/resources/admin-guide/proxy-protocol/

docker/ingress/build-nginx.sh
rewiko: Exactly the same issue for nginx build, and ubuntu didn't update openssl fast enough when you have a CVE.
I wonder if we can use libreSSL instead, seems much better without all the CVE of openssl.

rewiko:
Ready for review again, finally we use Ubuntu openssl lib.

jsravn:
@rewiko can you add CVE feed to our slack room?

rewiko:
subscribed https://www.mail-archive.com/openssl-announce@openssl.org/maillist.xml on the core slack room.

README.md
```diff
@@ -43,11 +42,26 @@ See the command line options with:
...
+ # Set default ssl path + name file without extension.  Feed expects two files: one ending in .crt (the CA) and the other in .key (the private key), for example:
+ -default-ssl-path=/etc/ssl/default-ssl/default-ssl
+ # ... will cause feed to look for /etc/ssl/default-ssl.crt and /etc/ssl/default.ssl.key
+```
+
+You can mount the `.key` and `.crt` though a Kubernetes Secret see [feed-ingress-deployment-ssl](examples/feed-ingress-deployment-ssl.yml).
+
 ### Gorb / IPVS Support
 
 feed has support for configuring IPVS via [gorb](https://github.com/sky-uk/gorb).
```
jsravn: Shouldn't we link the upstream project here?

nginx/nginx.tmpl
```diff
@@ -102,22 +99,28 @@ http {
...
+{{- $SSLPath := .SSLPath }}
+{{define "HTTPSConf"}}
+        # https://mozilla.github.io/server-side-tls/ssl-config-generator/ - Nginx, Modern Profile + TLSv1, TLSv1.1
+        ssl_certificate {{ . }}.crt;
+        ssl_certificate_key {{ . }}.key;
+        ssl_session_timeout 1d;
+        ssl_session_cache shared:SSL:50m;
+        ssl_session_tickets off;
+        ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
+        ssl_ciphers 'ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:!DSS';
```
jsravn: This doesn't look right - this is from the intermediate profile. It should be using AES only.

README.md
rewiko: For now, we are still releasing the binary on https://github.com/sky-uk/gorb and not the upstream. If we have an issue it's faster to fix in our repo and then try to merge on upstream.

nginx/nginx.tmpl
```diff
@@ -102,22 +99,28 @@ http {
...
+{{- $IngressPorts := .Ports }}
+{{- $SSLPath := .SSLPath }}
+{{define "HTTPSConf"}}
+        # https://mozilla.github.io/server-side-tls/ssl-config-generator/ - Nginx, Modern Profile + TLSv1, TLSv1.1
+        ssl_certificate {{ . }}.crt;
+        ssl_certificate_key {{ . }}.key;
+        ssl_session_timeout 1d;
+        ssl_session_cache shared:SSL:50m;
+        ssl_session_tickets off;
+        ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
```
jsravn: Can we use just TLSv1.2 until we run into problems? All modern clients that we're accepting should use this. (Not sure on this, welcome to opposing viewpoints). I'm mostly concerned about straying from the "modern profile" with our own customizations (might be some security issues with this we're not aware of).

rewiko: Depending if some client use IE lower than 11, well they should be deprecated. maybe we can only use TLSv1.2.

jsravn: I don't think old IE will work with this configuration anyways - not sure it supports the ciphers we want to use.

rewiko:
done

rewiko:
@jsravn Are we waiting on totahuanocotl to come back on Tuesday ?

jsravn:
No it's fine - I think you addressed his comments.

rewiko:
Ok so I don't have the right to force merge on this repo, so up to you to merge it :).

rewiko:
Now I have.

PR #163, opened 2017-11-29
Changing the path to python3 for the fake graceful nginx script

peterbale:
This is to change the path to python3.6 from using the absolute path to getting the dynamic path as it does not always live in /usr/bin/.


rewiko:
maybe only `#!/usr/bin/env python3` will make more versatile.

peterbale:
@rewiko happy to change to python3, anyone know of any libs the script uses which is 3.6 dependant?

PR #164, opened 2017-11-30
Gorb add HTTP retries using pester

peterbale:
Changing the http client gorb uses to pester which allows us to define retry settings. [Pester](https://github.com/sethgrid/pester) uses the built-in golang http client and adds a wrapper allowing us more functionality when making basic http requests.

Adding two new feed ingress flags:
* gorb-http-client-timeout (default: 2s)
* gorb-http-client-max-retries (default: 3)

### DIFF
--- a/cmd/feed-ingress/main.go
+++ b/cmd/feed-ingress/main.go
@@ -57,8 +57,6 @@ var (
 	gorbBackendHealthcheckInterval string
 	gorbBackendHealthcheckType     string
 	gorbInterfaceProcFsPath        string
-	gorbHTTPClientTimeout          time.Duration
-	gorbHTTPClientMaxRetries       int
 )
 
 const (
@@ -102,8 +100,6 @@ const (
 	defaultGorbInterfaceProcFsPath           = "/host-ipv4-proc/"
 	defaultGorbBackendHealthcheckInterval    = "1s"
 	defaultGorbBackendHealthcheckType        = "http"
-	defaultGorbHTTPClientTimeout             = time.Second * 2
-	defaultGorbHTTPClientMaxRetries          = 3
 )
 
 func init() {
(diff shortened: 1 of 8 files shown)


cmd/feed-ingress/main.go
```diff
@@ -292,6 +300,8 @@ func createIngressUpdaters() ([]controller.Updater, error) {
 			BackendHealthcheckInterval: gorbBackendHealthcheckInterval,
 			BackendHealthcheckType:     gorbBackendHealthcheckType,
 			InterfaceProcFsPath:        gorbInterfaceProcFsPath,
+			HTTPClientTimeout:          gorbHTTPClientTimeout,
+			HTTPClientMaxRetries:       gorbHTTPClientMaxRetries,
```
jsravn: I don't think we need to make the timeouts configurable. The ELB time outs are not configurable either.

gorb/gorb.go
```diff
@@ -122,9 +125,10 @@ func New(c *Config) (controller.Updater, error) {
 		backendDefinitions = append(backendDefinitions, backendDefinition)
 	}
 
-	var httpClient = &http.Client{
-		Timeout: time.Second * 5,
-	}
+	httpClient := pester.New()
+	httpClient.Timeout = c.HTTPClientTimeout
+	httpClient.MaxRetries = c.HTTPClientMaxRetries
+	httpClient.Backoff = pester.ExponentialBackoff
```
jsravn: May be better to not use a backoff, since we don't want this to retry for very long. Exponential would make more sense if we wanted to retry for a long period. I suppose if we used a max timeout of 60s or something exponential makes sense.

cmd/feed-ingress/main.go
peterbale: have removed the flags from feed-ingress and set the values in gorb

gorb/gorb.go
peterbale: have reverted back to the default backoff of 1s

PR #165, opened 2018-02-16
Replace govendor with dep

jsravn:
Also remove vendor directory completely from source control. Use `dep
ensure` instead.


jsravn:
This is prep before implementing merlin client.

PR #167, opened 2018-02-21
Add support for merlin frontend

jsravn:
To attach to merlin instances. Merlin is an IPVS based distributed load
balancer. See https://github.com/sky-uk/merlin.


jsravn:
Need to also handle attaching if previously attached - commit coming shortly.

controller/controller.go
```diff
@@ -89,10 +89,17 @@ func (c *controller) Start() error {
...
 	for _, u := range c.updaters {
 		if err := u.Start(); err != nil {
+			for _, started := range startedUpdaters {
+				if err := started.Stop(); err != nil {
+					log.Warnf("unable to stop %s: %v", u, err)
+				}
+			}
 			return fmt.Errorf("unable to start %v: %v", u, err)
 		}
+		startedUpdaters = append(startedUpdaters, u)
```
jsravn: This is just an additional safeguard to make sure we don't leave things dangling or in a broken state. See the test.

merlin/merlin.go
```diff
@@ -0,0 +1,170 @@
...
+	} else if err != nil {
+		return fmt.Errorf("unable to register with merlin: %v", err)
+	}
+
+	return nil
+}
+
+func (u *updater) detach() {
+	// drain
+	func() {
```
alangibson01: What's the purpose of declaring this block in a `func`?

merlin/netlink.go
```diff
@@ -0,0 +1,37 @@
...
+
+type nl interface {
+	addVIP(vipInterface, vip string) error
+	removeVIP(vipInterface, vip string) error
+}
+
+type nlImpl struct{}
+
+func (i *nlImpl) handleVIP(vipInterface, vip string, fn func(netlink.Link, *netlink.Addr) error) error {
+	lnk, err := netlink.LinkByName(vipInterface)
```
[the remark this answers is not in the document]
jsravn: Not sure what you mean? There isn't a handle to delete.

merlin/merlin.go
alangibson01: Ah yes, I remember you mentioning that the other week.

merlin/netlink.go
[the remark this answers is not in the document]
alangibson01: Misread the code, ignore this.

examples/feed-ingress-deployment-merlin.yml
```diff
@@ -0,0 +1,184 @@
...
+        # Interface to bind virtual IP to.
+        - -merlin-vip-interface=lo
+
+        # Each worker uses a full cpu, so scale up vertically on a box by increasing this value.
+        - -nginx-workers=1
+
+        # Connections*workers needs to be less than available ephemeral ports. Linux default is 60999-32768=28231.
+        - -nginx-worker-connections=20000
+
+        # Needs to be greater than the frontend idle timeout, see `aws_provisioning/kubernetes` for frontend value.
```
alangibson01: It's just occurred to me that referencing `aws_provisioning/kubernetes` makes no sense in the context of Feed as a standalone application. We should change this as well.

alangibson01: If it's referenced in other example files, we should change those too.

jsravn: We really should autogenerate these, or something.

jsravn: Anyways, I fixed it.

PR #170, opened 2018-03-13
Configure memory allocated to vhost stats module

alangibson01:
In some cases, this may need to be set to a value higher than the
default of 1 MiB, and is now configurable through the
`vhost-stats-shared-memory` flag. Default value is 1 MiB.


jsravn:
CHANGELOG.md will need to be coordinate with https://github.com/sky-uk/feed/pull/171 though.

PR #171, opened 2018-03-13
Add merlin health checks

jsravn:


jsravn:
Also forgot to update the CHANGELOG - will do that now.

PR #173, opened 2018-04-04
Attach to https in merlin

jsravn:
Also some small improvements:
* Expose the merlin request timeout, so it's easy to change if necessary
* Make deregistration more resilient, which could fail due to the long lived grpc connection getting killed. Instead, create and close connections on demand.

### DIFF
--- a/cmd/feed-ingress/main.go
+++ b/cmd/feed-ingress/main.go
@@ -60,9 +60,10 @@ var (
 	gorbBackendHealthcheckType     string
 	gorbInterfaceProcFsPath        string
 	merlinEndpoint                 string
+	merlinRequestTimeout           time.Duration
 	merlinServiceID                string
+	merlinHTTPSServiceID           string
 	merlinInstanceIP               string
-	merlinInstancePort             uint
 	merlinForwardMethod            string
 	merlinDrainDelay               time.Duration
 	merlinHealthUpThreshold        uint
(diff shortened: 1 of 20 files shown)


merlin/merlin_test.go
```diff
@@ -37,15 +37,20 @@ var _ = Describe("merlin", func() {
```
totahuanocotl: It seems like each test has checks for both http and https servers. 
I wonder if using the [ginkgo shared examples](http://onsi.github.io/ginkgo/#pattern-1-extract-a-function-that-defines-the-shared-its) would help? you could write the shared scenarios, and then have two contexts, one http and one https, and then just apply those checks. 
Give it a look and see if it would make sense.

cmd/feed-ingress/main.go
```diff
@@ -255,9 +256,11 @@ func init() {
 	flag.StringVar(&merlinEndpoint, "merlin-endpoint", "",
 		"Merlin gRPC endpoint to connect to. Expected format is scheme://authority/endpoint_name (see "+
 			"https://github.com/grpc/grpc/blob/master/doc/naming.md). Will load balance between all available servers.")
-	flag.StringVar(&merlinServiceID, "merlin-service-id", "", "Merlin virtual service ID to attach to.")
+	flag.DurationVar(&merlinRequestTimeout, "merlin-request-timeout", time.Second * 10,
+		"Timeout for any requests to merlin.")
+	flag.StringVar(&merlinServiceID, "merlin-service-id", "", "Merlin http virtual service ID to attach to.")
+	flag.StringVar(&merlinHTTPSServiceID, "merlin-https-service-id", "", "Merlin https virtual service ID to attach to.")
 	flag.StringVar(&merlinInstanceIP, "merlin-instance-ip", "", "Ingress IP to register with merlin")
-	flag.UintVar(&merlinInstancePort, "merlin-instance-port", 0, "Ingress port to register with merlin")
```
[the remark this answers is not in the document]
jsravn: I suppose, but the attaching wasn't working at all until the last version which got released just a couple days ago. So I don't think we need to do a major version release for this, it's part of the 1.9 which introduced merlin.

merlin/merlin.go
```diff
@@ -118,41 +124,71 @@ func (u *updater) registerWithMerlin() error {
 		Period:        ptypes.DurationProto(u.HealthPeriod),
 		Timeout:       ptypes.DurationProto(u.HealthTimeout),
 	}
+	if err := u.registerServer(server, "http"); err != nil {
```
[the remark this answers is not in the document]
jsravn: Yeah good idea.

merlin/merlin_test.go
jsravn: I'm not sure I follow - there is only a single assertion on the merlin client mock.

You mean stubbing the mocks? I'm hesitant to put that in a method because it obscures the test meaning.

totahuanocotl: nope, it is something similar as the `shared_examples` for the rspecs, where you define the expected cases, and then apply them to different targets. 
But i just realized there are only 3 tests like that, so not really worth it. Somehow I though there were many more.

PR #174, opened 2018-05-18
Add ingress loadbalancer status for elb and Merlin

peterbale:
This is to support feed updating the ingress status. By default this will be disabled, however, if you run `feed-ingress` using either the `elb` or `Merlin` updater and enable `update-ingress-status`, feed will update the status with the relevant loadbalancer data.

Part of: https://github.com/sky-uk/core-infrastructure/issues/2250


jsravn:
Is there a reason to make it optional? What would be the reason to disable it? I think it should just always update the status.

peterbale:
@jsravn I was seeing it as more of a feature flag but I don't think it's a big enough change to even need one anymore. I'll change it to just update the status and remove the flag.

cmd/feed-ingress/main.go
```diff
@@ -375,6 +395,11 @@ func createIngressUpdaters() ([]controller.Updater, error) {
 		updaters = append(updaters, gorbUpdater)
 
 	case "merlin":
+		if internalHostname == "" || externalHostname == "" {
+			return updaters, errors.New("Unable to update ingress for Merlin frontends. Missing flags " +
+				"'internal-hostname' and 'external-hostname'.")
```
mgreyo: Error message would report both variables as missing if one isn't set

cmd/feed-ingress/main.go
```diff
@@ -401,6 +426,17 @@ func createIngressUpdaters() ([]controller.Updater, error) {
...
+			KubernetesClient: controllerConfig.KubernetesClient,
+		}
+		merlinStatusUpdater, err := merlin_status.New(statusConfig)
+		if err != nil {
+			return updaters, err
+		}
+		updaters = append(updaters, merlinStatusUpdater)
+
 	default:
 		return nil, fmt.Errorf("invalid registration frontend type. Must be either gorb, elb, alb but was %s", registrationFrontendType)
```
mgreyo: or merlin

elb/status/status.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/sky-uk/feed/controller"
+	"github.com/sky-uk/feed/elb"
+	"github.com/sky-uk/feed/k8s"
+	"github.com/sky-uk/feed/util"
+
+	aws_elb "github.com/aws/aws-sdk-go/service/elb"
+)
+
+// Config for creating a new ingress controller.
```
mgreyo: status updater instead of ingress controller?

elb/status/status.go
```diff
@@ -0,0 +1,87 @@
...
+	labelValue       string
+	kubernetesClient k8s.Client
+	initialised      sync.Mutex
+}
+
+func (s *status) Start() error {
+	return nil
+}
+
+// Stop removes this instance from all the front end ELBs
```
mgreyo: comment can go

merlin/status/status.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/sky-uk/feed/util"
+	"k8s.io/client-go/pkg/api/v1"
+)
+
+const (
+	internalLabelValue = "internal"
+	externalLabelValue = "internet-facing"
+)
+
+// Config for creating a new ingress controller.
```
mgreyo: same as `elb/status/status.go`: updater instead of ingress controller?

merlin/status/status.go
```diff
@@ -0,0 +1,76 @@
...
+	lbs              map[string][]v1.LoadBalancerIngress
+	kubernetesClient k8s.Client
+	initialised      sync.Mutex
+}
+
+func (s *status) Start() error {
+	return nil
+}
+
+// Stop removes this instance from all the front end ELBs
```
mgreyo: same as `elb/status/status.go`: comment can go

util/status.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	return true
+}
+
+// SliceToStatus to convert a slice of strings to ingress loadbalancer objects.
+// Allows hostnames or ip addresses and sets the appropriate field.
+func SliceToStatus(endpoints []string) []v1.LoadBalancerIngress {
+	lbi := []v1.LoadBalancerIngress{}
+	for _, ep := range endpoints {
+		if net.ParseIP(ep) == nil {
```
mgreyo: perhaps remove the `== nil`, and swap the outcomes around to improve readability

util/status.go
```diff
@@ -0,0 +1,44 @@
...
+		} else {
+			lbi = append(lbi, v1.LoadBalancerIngress{IP: ep})
+		}
+	}
+
+	sort.SliceStable(lbi, func(a, b int) bool {
+		return lbi[a].Hostname < lbi[b].Hostname
+	})
+
+	sort.SliceStable(lbi, func(a, b int) bool {
```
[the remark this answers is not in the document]
totahuanocotl:
```
	sort.SliceStable(unsorted, func(i, j int) bool {
		if unsorted[i].y < unsorted[j].y {
			return true
		}
		if unsorted[i].y > unsorted[j].y {
			return false
		}
		return unsorted[i].x < unsorted[j].x
	})
```

util/status.go
```diff
@@ -12,11 +12,15 @@ func StatusUnchanged(existing, new []v1.LoadBalancerIngress) bool {
 	if len(existing) != len(new) {
 		return false
 	}
+
+	sortLoadBalancerIngress(existing)
```
totahuanocotl: I think this should also sort `new`, as there would be no guarantee that it comes pre-sorted

util/status.go
```diff
@@ -25,20 +29,26 @@ func StatusUnchanged(existing, new []v1.LoadBalancerIngress) bool {
...
 	}
 
-	sort.SliceStable(lbi, func(a, b int) bool {
-		return lbi[a].Hostname < lbi[b].Hostname
-	})
-
-	sort.SliceStable(lbi, func(a, b int) bool {
-		return lbi[a].IP < lbi[b].IP
-	})
+	sortLoadBalancerIngress(lbi)
```
totahuanocotl: should we just move the sort to StatusUnchanged? or keep both?

elb/status/status.go
```diff
@@ -42,14 +43,19 @@ type status struct {
 	awsElb           elb.ELB
 	labelValue       string
 	kubernetesClient k8s.Client
-	initialised      sync.Mutex
+	clusterFrontEnds map[string]elb.LoadBalancerDetails
```
totahuanocotl: would it make sense to have a consistent name for the LoadBalancer for both elb and merlin?
for some reason merlin has a simplified `lb`

util/status.go
```diff
@@ -25,20 +29,26 @@ func StatusUnchanged(existing, new []v1.LoadBalancerIngress) bool {
 func SliceToStatus(endpoints []string) []v1.LoadBalancerIngress {
```
totahuanocotl: could we use a different name for this method? every time i read Slice to Status, i can't think of what Slice means

elb/status/status.go
```diff
@@ -0,0 +1,90 @@
...
+		labelValue:       conf.LabelValue,
+		kubernetesClient: conf.KubernetesClient,
+	}, nil
+}
+
+type status struct {
+	awsElb           elb.ELB
+	labelValue       string
+	kubernetesClient k8s.Client
+	loadbalancers    map[string]elb.LoadBalancerDetails
```
totahuanocotl: Isn't load balancers a two word noun? Should be camel cased

peterbale: good spot :+1:

jsravn:
Having a quick look too...

jsravn:
CHANGELOG.md and README.md both need to be updated to reflect the new behavior and flags. Same with examples.

cmd/feed-ingress/main.go
```diff
@@ -375,6 +395,11 @@ func createIngressUpdaters() ([]controller.Updater, error) {
 		updaters = append(updaters, gorbUpdater)
 
 	case "merlin":
+		if internalHostname == "" || externalHostname == "" {
+			return updaters, errors.New("Unable to update ingress for Merlin frontends. Requires both " +
+				"'internal-hostname' and 'external-hostname' flags.")
+		}
```
jsravn: Why do we need both internal/external VIPs/hostnames? In m25 now we only have the internal one. I don't see why this is required either, if neither are provided just don't update the status.

Also, we should be able to query these values directly from merlin (list the virtual services and find their VIPs). So no need to require them as flags.

jsravn: Also, the names "internal" "external" should match the merlin flag names, which are "http-service" and "https-service". So what you want here is "httpServiceVIP", and "httpsServiceVIP". But like I said, we should just query these from merlin at startup.

cmd/feed-ingress/main.go
```diff
@@ -338,6 +347,17 @@ func createIngressUpdaters() ([]controller.Updater, error) {
 		}
 		updaters = append(updaters, elbUpdater)
 
+		statusConfig := elb_status.Config{
+			Region:           region,
+			LabelValue:       elbLabelValue,
+			KubernetesClient: controllerConfig.KubernetesClient,
```
jsravn: I think it would be better to pass the client into this function I think from main(). Rather than rely on the controller config, which may/may not change independently.

util/status.go
```diff
@@ -0,0 +1,44 @@
+package util
```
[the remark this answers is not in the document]
peterbale: that's fair, I could put it in a new `k8s/util` package?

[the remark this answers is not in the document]
jsravn: Still same, this shouldn't be in util.

elb/status/status.go
```diff
@@ -0,0 +1,90 @@
...
+			if err := s.kubernetesClient.UpdateIngressStatus(ingress.Ingress); err != nil {
+				log.Warn("Failed to update ingress status for %s: %e", ingress.Name, err)
+				updateFailed = true
+			}
+		}
+	}
+
+	if updateFailed {
+		return errors.New("failed to update all ingress statuses")
+	}
```
jsravn: We need to change how the controller handles errors from the updaters, since before we only ever had a single updater which could return errors. The problem with the controller currently is it aborts immediately on the first error. But if the status fails to update, we don't want to abort updating all ingresses, or vice versa. So we should change the controller to aggregate all errors.

elb/status/status.go
```diff
@@ -0,0 +1,90 @@
...
+			newStatus := util.GenerateLoadBalancerStatus([]string{lb.DNSName})
+
+			if util.StatusUnchanged(ingress.Ingress.Status.LoadBalancer.Ingress, newStatus) {
+				continue
+			}
+
+			ingress.Ingress.Status.LoadBalancer.Ingress = newStatus
+
+			if err := s.kubernetesClient.UpdateIngressStatus(ingress.Ingress); err != nil {
+				log.Warn("Failed to update ingress status for %s: %e", ingress.Name, err)
```
jsravn: This is going to be very spammy if there is a general problem updating ingress status (we have hundreds of ingresses in dev for instance). So we need to aggregate the errors up and produce a single log message ideally.

elb/status/status.go
```diff
@@ -0,0 +1,90 @@
...
+				updateFailed = true
+			}
+		}
+	}
+
+	if updateFailed {
+		return errors.New("failed to update all ingress statuses")
```
jsravn: The error should include the failed ingresses I think (abbreviated).

merlin/status/status.go
```diff
@@ -0,0 +1,79 @@
...
+	InternalHostname string
+	ExternalHostname string
+	KubernetesClient k8s.Client
+}
+
+// New creates a new Merlin frontend status updater.
+func New(conf Config) (controller.Updater, error) {
+	loadBalancers := make(map[string][]v1.LoadBalancerIngress)
+	loadBalancers[internalLabelValue] = util.GenerateLoadBalancerStatus([]string{conf.InternalHostname})
+	loadBalancers[externalLabelValue] = util.GenerateLoadBalancerStatus([]string{conf.ExternalHostname})
```
jsravn: The concept of "internal" and "external" ingress only makes sense for AWS. All our ingresses will be accessible via the external and internal VIPs, at least in the current setup.

cmd/feed-ingress/main.go
jsravn: Ok, I understand a bit better now, it is trying to match the AWS approach. But for M25, we don't have the concept of external or internal ingress like we do in AWS (where there is a separate external ELB and internal ELB). In M25 all ingresses are accessible through both the internal VIPs, as well as the external VIPs assigned to the palo alto. The internal setup could also have multiple VIPs, one for each service (http and https). I think we can derive the internal VIPs from merlin directly, but we can't do that for the "external" VIP. This is a tricky one. It seems like we need to provide that external VIP (or hostname, if that's needed) only, and then we need to use the frontend-scheme annotation to indicate the value to use so external-dns can assign the correct value. Need to think about it.

merlin/status/status.go
```diff
@@ -0,0 +1,79 @@
...
+		if lb, ok := s.loadBalancers[ingress.ELbScheme]; ok {
+			if util.StatusUnchanged(ingress.Ingress.Status.LoadBalancer.Ingress, lb) {
+				continue
+			}
+
+			ingress.Ingress.Status.LoadBalancer.Ingress = lb
+
+			if err := s.kubernetesClient.UpdateIngressStatus(ingress.Ingress); err != nil {
+				log.Warn("Failed to update ingress status for %s: %e", ingress.Name, err)
+				updateFailed = true
```
jsravn: Same as for the ELB updater, it will be spammy.

We should also probably consolidate this code into a shared lib since it's almost exactly the same in both ELB and merlin status updaters.

cmd/feed-ingress/main.go
jsravn: Does external-dns support creating explicit separate DNS entries for an ingress, via annotation or something? For an ingress using merlin, I think the load balancer status should contain the ingress VIPs. Any other hostname/vip is custom and depends entirely on the local setup, so I'm thinking this should be handled with a different mechanism. Although using the annotation to override that VIP is one approach (but will mean we likely need separate ingress for internal and external which isn't really necessary like it is with AWS).

peterbale: I was trying to use the same logic that was used in `feed-dns` (where the VIP's are defined as flags). Surely the `http` and `https` services run on the same IP(s) so surely we can just use the `merlinInstanceIP` as the only 'internal' ingress point? Once 'external' ingress has been added we can update the merlin status updater accordingly.

peterbale:
@jsravn I have made a few changes which should address your comments. As for the CHANGELOG, I was unsure if it would be `v1.9.4` or `1.10.0` as it's a moderately sizeable change?

cmd/feed-ingress/main.go
peterbale: I have added a new `merlinInternetFacingVIP` flag which for now will just be metadata for the frontend ip address (will only be used for setting DNS and general information about the status). Going forward, we will probably want to grab both this value and `merlinVIP` directly from Merlin.

elb/status/status.go
peterbale: I have changed the shared function to aggregate ingress names that fail and then print out at the end (assuming at least one failed)

elb/status/status.go
peterbale: Addressed with aggregated failed ingresses.

elb/status/status.go
peterbale: Addressed with aggregated failed ingresses.

merlin/status/status.go
peterbale: Addressed with new `merlinInternetFacingVIP` flag which allows setting the external ip address (currently just used as metadata for setting the status). At a later date we may want to query Merlin for both this VIP and the `merlinVIP`.

merlin/status/status.go
peterbale: Addressed with aggregated failed ingresses.

util/status.go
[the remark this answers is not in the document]
peterbale: have moved into `k8s/status` which is now a shared library for setting the status (which in turn uses the k8s client)

jsravn:
Definitely 1.10.

PR #175, opened 2018-06-04
Add 'empty' or 'None' service name check while generating upstream entries

ashutoshg:
https://github.com/sky-uk/core-infrastructure/issues/2421

### DIFF
--- a/nginx/nginx.go
+++ b/nginx/nginx.go
@@ -402,7 +402,7 @@ func createUpstreamEntries(entries controller.IngressEntries) []*upstream {
 	idToUpstream := make(map[string]*upstream)
 
 	for _, ingressEntry := range entries {
-		if validUpstreamEntry(ingressEntry)  {
+		if validUpstreamEntry(ingressEntry) {
 			upstream := &upstream{
 				ID:             upstreamID(ingressEntry),
 				Server:         fmt.Sprintf("%s:%d", ingressEntry.ServiceAddress, ingressEntry.ServicePort),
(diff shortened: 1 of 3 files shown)


nginx/nginx.go
```diff
@@ -402,12 +402,14 @@ func createUpstreamEntries(entries controller.IngressEntries) []*upstream {
 	idToUpstream := make(map[string]*upstream)
 
 	for _, ingressEntry := range entries {
-		upstream := &upstream{
-			ID:             upstreamID(ingressEntry),
-			Server:         fmt.Sprintf("%s:%d", ingressEntry.ServiceAddress, ingressEntry.ServicePort),
-			MaxConnections: ingressEntry.BackendMaxConnections,
+		if validUpstreamEntry(ingressEntry) {
```
peterbale: I think we would want some sort of log for when we get an invalid ingress. We could maybe aggregate the invalid ingresses and then log them as a warn?

nginx/nginx.go
```diff
@@ -419,6 +421,10 @@ func createUpstreamEntries(entries controller.IngressEntries) []*upstream {
 	return sortedUpstreams
 }
 
+func validUpstreamEntry(e controller.IngressEntry) bool {
```
peterbale: Is this validation only applicable for the `nginx` updater? We could maybe have this validation done at the client or controller level otherwise we are throwing invalid ingresses around all the time. Maybe the `GetIngresses()` function in the client library might be better?

peterbale: Centralising this check makes logging the aggregated invalid ingresses easier too :smile:

ashutoshg: I thought about that option as well but as the issues are only related to while creating upstream config I kept it related to upstream creation only.

nginx/nginx.go
ashutoshg: Its nice Idea to log, I am not sure aggregating though, as we do not expect loads of them.

ashutoshg:
closing this PR in favour of https://github.com/sky-uk/feed/pull/177

PR #176, opened 2018-06-04
Adding internal and internet facing hostnames

peterbale:
Due to merlin using a cname for all DNS entries rather than a single ip
address, we need to supply both the internal and internet facing
hostnames as the ingress status. This will then allow external-dns to
create a cname from the ingress hostname to this relevant 'ingress'
hostname.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 .idea
 *.iml
 *.swp
+*.ipr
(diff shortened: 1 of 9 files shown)


peterbale:
Waiting for #177 to be merged in and then this PR is ready for final approval and merge to master

PR #177, opened 2018-06-04
Added ignore for 'None' cluster IP while updating nginx config

ashutoshg:
fixes: https://github.com/sky-uk/core-infrastructure/issues/2421

### DIFF
--- a/controller/ingress_entry.go
+++ b/controller/ingress_entry.go
@@ -50,6 +50,9 @@ func (e IngressEntry) validate() error {
 	if e.ServiceAddress == "" {
 		return errors.New("missing service address")
 	}
+	if e.ServiceAddress == "None" {
+		return errors.New("missing service address")
+	}
 	if e.ServicePort == 0 {
 		return errors.New("missing service port")
 	}
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 .idea
 *.iml
 *.swp
+*.ipr
+*.iws
 build
 vendor
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -351,6 +351,12 @@ func TestUpdaterIsUpdatedOnK8sUpdates(t *testing.T) {
 			nil,
 		},
 		{
+			"ingress with 'None' as service IP",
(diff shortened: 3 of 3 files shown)


controller/ingress_entry.go
```diff
@@ -50,6 +50,9 @@ func (e IngressEntry) validate() error {
 	if e.ServiceAddress == "" {
```
peterbale: I think it might be cleaner to have the check here as we treat an empty string and `None` the same

peterbale: Also, will we see the same issue if `Host` gets set to `None`?

ashutoshg: Only cluster IP gets set to 'None', I think from error point of view its good to have information about the None, address I propose may be changing this to have message as `Service address is set to None ignoring entry` or on similar lines, may be we can include actual entry in logging on other places as well.

Let me know your thoughts.

peterbale: Yeah I think updating the log message with something like that and leaving it as two checks sounds a good idea :+1:

ashutoshg: cool that's changed now.

controller/ingress_entry.go
```diff
@@ -50,6 +50,9 @@ func (e IngressEntry) validate() error {
...
 		return errors.New("missing service address")
 	}
+	if e.ServiceAddress == "None" {
+		return errors.New("Service address is set to 'None' ignoring entry")
```
peterbale: The error is handled upstream so I don't think we need the 'ignoring entry' part. Also, could we have the log all lower case too for consistency when printing (as this will be appended to an error) so we will just have `service address is set to 'None'`. After that I think this is good to go.

peterbale: Also, when releasing can you update this PR to merge into the `add-merlin-hostnames` branch so we can then have one PR which is merging into master with one changelog and one release.

ashutoshg: cool :+1:

ashutoshg:
Addressed comments and changed the branch to be merge to `add-merlin-hostnames` as mentioned by @peterbale

PR #180, opened 2018-08-03
Expose status updater errors

### LINKED ISSUE #179: Improve logging on ingress status update failure
Should we consider logging the error here? At present we only log that it has failed but not the reason why.

https://github.com/sky-uk/feed/blob/288c881ee722f1a3547bdb162ec62eb9f6f8b824/k8s/status/status.go#L40-L42

peterbale:
This is to return back any errors recieved when trying to set an ingress
status rather than printing the names of the failed ingresses.

Closes #179

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v1.10.3
+* Expose status updater error logs rather than printing a list of failed ingresses
+
 # v1.10.2
 * Bug fix for k8s/status updater where feed would exit the update loop if any ingress was 'unchanged'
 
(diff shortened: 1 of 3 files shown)


peterbale:
I can see what you're saying around the use of multierror. Initially I was trying to find some way of deduping the errors, but it's not really possible. The error includes the full path it was trying to hit, so it will include the namespace and ingress etc, hence why it's so hard to dedupe. I'll update it to use a slice of errors instead as I do agree, it's pretty much the same without using the lib.

PR #181, opened 2018-09-05
Add OpenTracing support

lw346:
See https://github.com/opentracing-contrib/nginx-opentracing

The idea here is to allow command line options to enable OpenTracing with standardised predictable operation names.

As OpenTracing is just a standard set of APIs, this doesn't force anyone to use a particular vendor implementation on the backend, but as Jaeger is part of the CNCF, we package it with the build as a reasonable "default" implementation.

We explicitly don't trace the default location, nor the status or metrics endpoints to reduce the amount of unnecessary traces being forwarded on to the backend.

If we are happy with the functionality of the change, we should update the changelog for the next version to cut.


nginx/nginx_test.go
```diff
@@ -859,6 +883,60 @@ func TestNginxIngressEntries(t *testing.T) {
...
+					Path:           "",
+					ServiceAddress: "service",
+					ServicePort:    9090,
+				},
+			},
+			nil,
+			[]string{
+				"opentracing_operation_name '$server_name /';",
+			},
+		},
```
totahuanocotl: Should we add a check for none of the opentracing directives if they are not enabled?

lw346: Already checked by "Check full ingress entry works"

rewiko:
We will need to check on dev because feed might use more memory.

PR #185, opened 2018-11-06
Enable overriding proxy buffer values

supreethrao:
- The default proxy buffer size has been changed to 16k
- The default proxy buffer blocks has been changed to 4 16k
- Both of the above values can be overridden with the relevant
annotations
- There is a max permissible value and if the annotated value exceeds
that, it will be reset to the max permissible value for both of the
above fields.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # v1.11.3
 * Enable overriding proxy buffer values. Defaults to `proxy_buffer_size 16k` and `proxy_buffers 4 16k`
 Can be overridden with relevant annotations
-* The values overridden by the annotations are capped at a permissible max `proxy_buffer_size 32k` and `proxy_buffers 8 32k`  
+* The values overridden by the annotations are capped at a permissible max `proxy_buffer_size 32k` and `proxy_buffers 8 32k`
+  
 # v1.11.2
 * Update nginx-opentracing version for bug fix to proxy headers
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v1.11.3
+* Enable overriding proxy buffer values. Defaults to `proxy_buffer_size 16k` and `proxy_buffers 4 16k`
+Can be overridden with relevant annotations
+* The values overridden by the annotations are capped at a permissible max `proxy_buffer_size 32k` and `proxy_buffers 8 32k`  
 # v1.11.2
 * Update nginx-opentracing version for bug fix to proxy headers
 
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -21,19 +21,26 @@ import (
 	"k8s.io/client-go/pkg/api/v1"
 )
 
-const ingressAllowAnnotation = "sky.uk/allow"
-const frontendSchemeAnnotation = "sky.uk/frontend-scheme"
+const (
+	ingressAllowAnnotation   = "sky.uk/allow"
+	frontendSchemeAnnotation = "sky.uk/frontend-scheme"
 
-// Deprecated: retained to maintain backwards compatibility.
-const frontendElbSchemeAnnotation = "sky.uk/frontend-elb-scheme"
-const stripPathAnnotation = "sky.uk/strip-path"
+	// Deprecated: retained to maintain backwards compatibility.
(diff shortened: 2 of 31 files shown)


controller/controller.go
```diff
@@ -220,6 +231,24 @@ func (c *controller) updateIngresses() error {
 						entry.BackendMaxConnections = tmp
 					}
 
+					if proxyBufferSizeString, ok := ingress.Annotations[proxyBufferSizeAnnotation]; ok {
+						tmp, _ := strconv.Atoi(proxyBufferSizeString)
+						entry.ProxyBufferSize = tmp
+						if tmp > maxAllowedProxyBufferSize {
+							log.Warnf("ProxyBufferSize %d has a value which exceeds the max permissible value. Resetting to: %d.", tmp, maxAllowedProxyBufferSize)
```
mgreyo: I don't think resetting is the correct word here. It should tell the user that the value they have set exceeds `maxAllowedProxyBufferSize` and that it will use that value instead.

controller/ingress_entry.go
```diff
@@ -40,6 +40,10 @@ type IngressEntry struct {
 	CreationTimestamp time.Time
 	// Ingress resource
 	Ingress *v1beta1.Ingress
+	//
```
mgreyo: `golint` complains for a reason - exported properties should have a simple explanation.

controller/controller.go
supreethrao: Agreed. Changed as per suggestion

controller/controller_test.go
```diff
@@ -490,6 +510,82 @@ func TestUpdaterIsUpdatedOnK8sUpdates(t *testing.T) {
...
+				Host:                  ingressHost,
+				Path:                  ingressPath,
+				ServiceAddress:        serviceIP,
+				ServicePort:           ingressSvcPort,
+				LbScheme:              "",
+				Allow:                 []string{},
+				StripPaths:            false,
+				BackendTimeoutSeconds: 10,
+				BackendMaxConnections: defaultMaxConnections,
+				ProxyBufferSize:       32,
```
peterbale: Is this not the max? Can we set it to a value greater than the max instead? Additionally, it might be worth using the `maxAllowedProxyBuffer*` constants here so that it doesn't break the tests if we raise/drop them.

merlin/status/status.go
```diff
@@ -6,9 +6,8 @@ package status
 import (
 	"github.com/sky-uk/feed/controller"
 	"github.com/sky-uk/feed/k8s"
-	"k8s.io/client-go/pkg/api/v1"
-
 	k8s_status "github.com/sky-uk/feed/k8s/status"
+	"k8s.io/client-go/pkg/api/v1"
```
peterbale: It was correct before.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v1.11.3
+* Enable overriding proxy buffer values. Defaults to `proxy_buffer_size 16k` and `proxy_buffers 4 16k`
+Can be overridden with relevant annotations
+* The values overridden by the annotations are capped at a permissible max `proxy_buffer_size 32k` and `proxy_buffers 8 32k`
```
peterbale: New line

controller/controller_test.go
supreethrao: Not sure I follow :( The ingress entry is our expectation. The test is when
 ```
proxyBufferSizeAnnotation: "64",
proxyBufferBlocksAnnotation: "12",
```  is the annotation, we will use 32 and 8 instead

File: controller/controller_test.go [hunk 6]
```diff
@@ -501,7 +597,13 @@ func TestUpdaterIsUpdatedOnK8sUpdates(t *testing.T) {
 		// setup clients
 		client := new(fake.FakeClient)
 		updater := new(fakeUpdater)
-		controller := newController(updater, client)
+
+		config := test.config
```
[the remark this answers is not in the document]
supreethrao: It is for the below 2 lines. We could say `test.config.KubernetesClient = client` if that's what you prefer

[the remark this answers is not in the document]
peterbale: It’s a test, it’s fine

supreethrao:
Thanks @peterbale

Issue #186, opened 2018-11-07
Configure feed ingress to handle requests with large headers and/or body 


supreethrao:
While working on the code changes to handle large response headers, @rewiko and I noticed that feed ingress rejected requests with large headers as well. We propose to increase the current default for feed-ingress to be able to handle such requests. The suggestion is to change the following configuration properties for values as mentioned below.

```
client_header_buffer_size       16k;
large_client_header_buffers 4 16k;
client_body_buffer_size 16k;
```

* The above mentioned defaults can be overridden via input arguments
* The configuration is for nginx http configuration which applies to all proxied servers. It cannot be individually configured per proxied server.

supreethrao:
This feature should be available in `v1.12.0`

PR #187, opened 2018-11-08
Support handling of large client requests

supreethrao:
Connects https://github.com/sky-uk/feed/issues/186

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,8 @@ Can be overridden with relevant annotations
 * The values overridden by the annotations are capped at a permissible max `proxy_buffer_size 32k` and `proxy_buffers 8 32k`
 * Supports handling of large client requests (header and body). It can be enabled by passing relevant command line args.
 
-# v1.11.3
-* Enable overriding proxy buffer values. Defaults to `proxy_buffer_size 16k` and `proxy_buffers 4 16k`
-Can be overridden with relevant annotations
-* The values overridden by the annotations are capped at a permissible max `proxy_buffer_size 32k` and `proxy_buffers 8 32k`
+# v1.11.4
+* DO NOT USE THIS VERSION
   
 # v1.11.2
 * Update nginx-opentracing version for bug fix to proxy headers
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# v1.12.0
+* Enable overriding proxy buffer values. Defaults to `proxy_buffer_size 16k` and `proxy_buffers 4 16k`
```
rewiko: 2 lines duplicate with v1.11.3

supreethrao: Yeh, I did not mean to release the `proxy_buffer` changes as a patch version. I am hoping I could include the changes from this and release them together as `1.12.0`. We then can deprecate `1.11.4` if that's an option. Although changes in `1.11.4` is not a breaking change but it is a new functionality and it makes sense to release them as a minor release.

nginx/nginx_test.go
```diff
@@ -367,7 +424,7 @@ func TestNginxConfig(t *testing.T) {
 
 		for _, expected := range test.expectedSettings {
 			if strings.HasPrefix(expected, "!") {
-				assert.NotContains(configContents, expected, "%s\nshould not contain setting", test.name)
+				assert.NotContains(configContents, expected[1:], "%s\nshould not contain setting", test.name)
```
supreethrao: I don't think this assertion ever worked because, the expected string contained `!` which can never be in the nginx conf and hence was passing incorrectly. The expected string should be sliced to remove the `!` character.

CHANGELOG.md
supreethrao: It's actually `1.11.4`. Is it ok if I remove the change notes for that release or mark it as depcrecated or donot use ? We can use `1.12.0` as the official release which supports large requests and responses.

cmd/feed-ingress/main.go
```diff
@@ -219,6 +219,9 @@ func init() {
 		"Path to OpenTracing plugin on disk (eg. /usr/local/lib/libjaegertracing_plugin.so)")
 	flag.StringVar(&nginxOpenTracingConfigPath, "nginx-opentracing-config-path", defaultNginxOpenTracingConfigPath,
 		"Path to OpenTracing config on disk (eg. /etc/jaeger-nginx-config.json)")
+	flag.IntVar(&nginxConfig.ClientHeaderBufferSize, "nginx-client-header-buffer-size-in-kb", 0, "Sets buffer size for reading client request header")
```
rewiko: I think we should set the default value equals to ProxyBufferBlocks, because they are both tighly coupled. "16k"

supreethrao:
Thanks @rewiko

PR #190, opened 2018-11-28
Patch comments according to best practices in CONTRIBUTING.md

mullikine:
Use Codelingo to automatically fix function comments following the
[effective go guidelines](https://golang.org/doc/effective_go.html#commentary) in [CONTRIBUTING.md](CONTRIBUTING.md).

This patch was generated by running the Codelingo Rewrite Flow over the "[comment-first-word-as-subject](https://github.com/codelingo/codelingo/blob/master/tenets/codelingo/effective-go/comment-first-word-as-subject/codelingo.yaml)" Tenet. Note: the same Tenet can be used to automate PR reviews and generate guideline docs. Install CodeLingo: https://github.com/apps/codelingo

### DIFF
--- a/dns/r53/client.go
+++ b/dns/r53/client.go
@@ -42,7 +42,7 @@ func New(hostedZone string, retries int) Route53Client {
 	}
 }
 
-// GetHostedZone gets the domain for the hosted zone
+// GetHostedZoneDomain gets the domain for the hosted zone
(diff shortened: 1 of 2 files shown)


totahuanocotl:
@mullikine thanks for the patch. 
We will look into codelingo.

PR #193, opened 2019-01-21
Fix duplicate path when path is not specified

yorg1st:
This fixes the issue where a duplicate ingress arises between a
path that has not been specified and a path of "/"

Connects https://github.com/sky-uk/core-infrastructure/issues/3686


yorg1st:
Fixes https://github.com/sky-uk/core-infrastructure/issues/3686

PR #194, opened 2019-01-31
Remove unnecessary config reload after start up

domleb:
We start nginx after generating the config so we don't need to signal a
reload.  This was causing nginx to reload the config 5 minutes after
every startup which we know can be disruptive and is unnecessary.

Set the 'initial update' flag after starting nginx as this will ensure we
don't return true on the health endpoing before nginx is running.
Previously we set this flag just before starting nginx so it's
possible we were reporting healthy before the process was running.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+# v1.12.2
+* Formatting changes
+  https://github.com/sky-uk/feed/pull/190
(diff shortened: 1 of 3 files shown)


domleb:
Fixes https://github.com/sky-uk/core-infrastructure/issues/3741

PR #195, opened 2019-01-31
Support multiple feed-ingress controllers per environment

domleb:
For https://github.com/sky-uk/core-infrastructure/issues/3414

This is a breaking change for `feed-ingress` to support creating multiple ingresses per environment.  This could be used for testing `feed-ingress` itself and for splitting traffic into different isolated ingresses per environment each with its own set of ELBs.

The change has been applied to `feed-ingress` but not `feed-dns`.  However the shared modules are backwards compatible and support both.

This change now requires the `--ingress-controller-name` argument to be passed to the `feed-ingress` container and the target ELBs must be tagged with `sky.uk/KubernetesClusterIngressControllerName`.  It won't attached to an ELB if it can't find the tag and thus it's a breaking change.  This was done to avoid the situation where a test instance could attach to the main ELBs for an environment if they haven't been tagged yet.

The `sky.uk/KubernetesClusterIngressControllerName` tag must match with `--ingress-controller-name` for `feed-ingress` to attach to the ELBs.

Ingress resources should be tagged with the `kubernetes.io/ingress.class` annotation that should match the name of the feed-ingress instance (specified with the `--ingress-controller-name` flag). Feed will adopt those ingress resources whose annotations match.

For migration the (new, deprecated) `--include-unnamed-ingresses` flag may be supplied, in which case feed-ingress will adopt all ingresses without the annotation, _in addition to those that match the controller name_.

required by https://github.com/sky-uk/core-infrastructure/issues/3414


domleb:
Connects https://github.com/sky-uk/core-infrastructure/issues/3414

cmd/feed-ingress/main.go
```diff
@@ -40,6 +40,8 @@ var (
 	pushgatewayIntervalSeconds     int
 	pushgatewayLabels              cmd.KeyValues
 	controllerConfig               controller.Config
+	mainIngress                    bool
```
peterbale: I think it might be clearer to call this variable `defaultIngress` and have `ingressName` default to `main` so that as a user, by default this feed instance can act as the default controller and attach to the main elb pair.

elb/elb.go
```diff
@@ -18,30 +18,38 @@ import (
...
+	if frontendTagValue == "" {
+		return nil, fmt.Errorf("unable to create ELB updater: missing value for the tag %v", ElbTag)
+	}
+	if ingressNameTagValue == "" {
+		return nil, fmt.Errorf("unable to create ELB updater: missing value for the tag %v", IngressNameTag)
 	}
+
 	initMetrics()
-	log.Infof("ELB Front end region: %s cluster: %s expected frontends: %d", region, labelValue, expectedNumber)
+	log.Infof("ELB Front end region: %s cluster: %s expected frontends: %d", region, frontendTagValue, expectedNumber)
```
peterbale: It would be good to include the ingress name tag here too

elb/elb.go
```diff
@@ -190,19 +211,34 @@ func FindFrontEndElbs(awsElb ELB, labelValue string) (map[string]LoadBalancerDet
...
 		// todo cb error out if we already have an internal or public facing elb
-		for _, description := range output.TagDescriptions {
-			for _, tag := range description.Tags {
-				if *tag.Key == ElbTag && *tag.Value == labelValue {
-					log.Infof("Found frontend elb %s", *description.LoadBalancerName)
-					lb := allLbs[*description.LoadBalancerName]
-					clusterFrontEnds[lb.Scheme] = lb
-				}
+		for _, elbDescription := range output.TagDescriptions {
+
```
peterbale: New line can be removed

elb/elb.go
```diff
@@ -190,19 +211,34 @@ func FindFrontEndElbs(awsElb ELB, labelValue string) (map[string]LoadBalancerDet
...
+				lb := allLbs[*elbDescription.LoadBalancerName]
+				clusterFrontEnds[lb.Scheme] = lb
 			}
 		}
 	}
 	return clusterFrontEnds, nil
 }
 
+func tagsDoMatch(elbTags []*aws_elb.Tag, tagsToMatch map[string]string) bool {
+
```
peterbale: New line can be removed

domleb:
> Rather than having it a global option I think this should be to support multiple elb ingresses which means some of the code can come out of the controller and into the elb code.

As discussed the ELB updater, nginx updater and ingress updater all have to apply these changes.  Actually the nginx updater doesn't need to do anything special as the controller filters the ingresses but if it didn't then the nginx updater would have to.  So I do think it's a global change as all parts have to work together for this to work.

cmd/feed-ingress/main.go
domleb: > have `ingressName` default to `main` so that as a user, by default this feed instance can act as the default controller and attach to the main elb pair.

I thought about this but decided it was safer to not default to main.  The reason is that if a test instance is deployed without this configuration it would then attach to the main ELBs and potentially break the main ingress.  What do you think?

controller/controller.go
```diff
@@ -275,6 +288,19 @@ func (c *controller) updateIngresses() error {
...
+	isValid := false
+
+	if ingressName, ok := ingress.Annotations[ingressNameAnnotation]; ok {
+		isValid = ingressName == c.ingressName
+	} else if c.mainIngress {
+		isValid = true
+	}
+
+	return isValid
+}
```
totahuanocotl: I think this could be simplified as
```suggestion
        if ingressName, ok := ingress.Annotations[ingressNameAnnotation]; ok {
		return ingressName == c.ingressName
	}
        return c.mainIngress
```

elb/elb.go
```diff
@@ -190,19 +211,31 @@ func FindFrontEndElbs(awsElb ELB, labelValue string) (map[string]LoadBalancerDet
...
+		log.Debugf("Checking for %s tag set to %s", name, value)
+		for _, elb := range elbTags {
+			if name == *elb.Key && value == *elb.Value {
+				matches++
+			}
+		}
+	}
+
+	return matches == len(tagsToMatch)
+}
```
totahuanocotl: I think we should approach this differently, so that we don't loop over the elbTags for each tag to Match:
1. convert the elbTags to a map first (1 loop), 
2. iterate over the tagsToMatch (1 loop) 
   1. if the tag to match key and value are not on the elb tags map, return false.
3. return true.

totahuanocotl:
I added a comment above but think it may get lost: https://github.com/sky-uk/feed/pull/195#discussion_r257153139

supreethrao:
> I added a comment above but think it may get lost: [#195 (comment)](https://github.com/sky-uk/feed/pull/195#discussion_r257153139)

FoMO :D

cmd/feed-ingress/main.go
domleb: > couldn't we have an ingress without a name? and avoid this main boolean?
I am just unsure as of why do we need it. If we want to maintain backwards compatibility, a nameless feed would be a way to have it, and if you want specific ingresses, you name them

The three of us discussed the two approaches:
1. The current implementation of a 'main' feed that handles ingresses that don't have an ingress name specified for backwards compatibility
- requires the main elbs to be tagged and feed-ingress can only attach if it's given the tag value
- this is a safety feature to ensure that test feed-ingress pods that haven't been configured don't attach to the main elbs
- requires an extra bool argument so that the main feed-ingress instance know to handle ingresses without a name, for backwards compatibility
2. An ingress that doesn't have a name that will be the main ingress
- don't add the ingress-name tag to the main elb and never add ingress-name to ingresses that are for the main / default elb
- doesn't require the main-ingress bool to handle two cases
- may need to update the logic that attaches to the elb to handle the no tag case
- potential risk that test pods without ingress-name specified could attach to the main elb
- could mitigate this by e.g. requiring the ingress-name arg and checking that it's not the same as a default randomly generated string (although, there's still a chance that someone will set a required arg to an empty string if they are just playing around :)

totahuanocotl: >* don't add the ingress-name tag to the main elb ...
>* may need to update the logic that attaches to the elb to handle the no tag case

we also said that we could intentionally add the tag our current elb with an empty name, so that the logic needs no change. I think an empty string is valid (unnamed). 

> * (although, there's still a chance that someone will set a required arg to an empty string if they are just playing around :)

given that the params should be given as either  `--main-ingress` (option 1) or `--ingress-name ""` (option 2), I'd say that there are more chances of people accidentally using --main-ingress.

domleb: `--main-ingress` defaults to false so this would need to be given as `--main-ingress=true` (or one of the other variants of true) so you would have to intentionally being setting it.  Still not fool proof so not ideal.

totahuanocotl: I think flags are not really the mechanism we would use to prevent one feed-ingress attaching to an incorrect elb by miss-configuration. It would probably be more on the side of "does this instance of feed has credentials with permissions to attach itself to this elb?"

domleb: In our case, test elbs can't actually attach to the main elb because they get a different role.  But that's just how we've done it (recently) and not provided by feed.

totahuanocotl: Ah great! In my opinion this should be that safety mechanism :)

supreethrao: I am of the opinion that the boolean field may not be necessary. This is because we are anyways advertising this as a breaking change, which means we can lose backwards compatibility. I would suggest for the feed ingress to process only those entries with the name annotation matching the name of the ingress and ignore anything which does not contain the name. 

It would still be a seamless change for the application teams at least in our case because we can make the change in k8plugin to have a name defaulted to the main ingress (which is the equivalent of the current breed of ingress we have). There can be a cut-off date before which all the teams should specify the ingress name. 

I am not fully convinced that we need to protect the test app getting attached to the main ingress/ELB. Such a mechanism does not exist now and also there's nothing stopping from the test attaching to the main ingress by specifying the name of the main ingress. Such mistakes can be caught via monitoring and alerting and am not convinced we need a boolean flag to protect. 

I would simply make the name mandatory for the ingress to start and a corresponding ELB with a tag matching the ingress name, without which the ingress should not get to running state.

controller/controller.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/sky-uk/feed/k8s"
 	"github.com/sky-uk/feed/util"
 	v1 "k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/pkg/apis/extensions/v1beta1"
```
supreethrao: Might need to use `apps/v1` package for deployments now that we are on 1.10 k8s.

controller/controller.go
```diff
@@ -172,7 +181,12 @@ func (c *controller) updateIngresses() error {
 
 				serviceName := serviceName{namespace: ingress.Namespace, name: path.Backend.ServiceName}
 
-				if address := serviceMap[serviceName]; address != "" {
+				if address := serviceMap[serviceName]; address == "" {
+					skipped = append(skipped, fmt.Sprintf("%s/%s (service doesn't exist)", ingress.Namespace, ingress.Name))
+				} else if !c.validIngressName(ingress) {
```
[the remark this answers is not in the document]
supreethrao: Simplifies this logic as per my above comment as it is mandatory for the ingress entries to have a name, else they will be ignored. No need to have an additional check on the boolean flag to allow these anonymous entries.

peterbale:
There seem to be some outstanding questions around the implementation of this and I think we should come to an agreement together before making further changes to the code. I'll assign it back to @domleb so we can discuss and agree a solution.

domleb:
Removed from review as this is in the backlog at the moment and not being worked on

howardburgess:
Rebased and fixed conflicts

controller/controller_test.go
```diff
@@ -270,508 +277,815 @@ func TestUnhealthyIfUpdaterFails(t *testing.T) {
...
+		createIngressesFixture(ingressNamespace, ingressHost, ingressSvcName, ingressSvcPort, map[string]string{
+			ingressAllowAnnotation:      "",
+			stripPathAnnotation:         "false",
+			backendTimeoutSeconds:       "10",
+			frontendElbSchemeAnnotation: "internal",
+			ingressNameAnnotation:       defaultIngressName,
+		}, ingressPath),
+		nil,
+		createNamespaceFixture(ingressNamespace, map[string]string{"team": "otherteam"}),
+		nil,
```
howardburgess: Expect no entries if the ingress is in a namespace without the matching label selector

controller/controller_test.go
```diff
@@ -270,508 +277,815 @@ func TestUnhealthyIfUpdaterFails(t *testing.T) {
...
+		"ingress is in a namespace that matches the namespace selector",
+		createIngressesFixture(ingressNamespace, ingressHost, ingressSvcName, ingressSvcPort, map[string]string{
+			ingressAllowAnnotation:      "",
+			stripPathAnnotation:         "false",
+			backendTimeoutSeconds:       "10",
+			frontendElbSchemeAnnotation: "internal",
+			ingressNameAnnotation:       defaultIngressName,
+		}, ingressPath),
+		createDefaultServices(),
+		createNamespaceFixture(ingressNamespace, map[string]string{"team": "theteam"}),
```
howardburgess: The ingress is in a namespace with the desired `team=theteam` label.

controller/controller_test.go
```diff
@@ -270,508 +277,815 @@ func TestUnhealthyIfUpdaterFails(t *testing.T) {
...
+	client.On("GetAllIngresses").Return(ingresses, nil)
+}
 
-		updater.On("Start").Return(nil)
-		updater.On("Stop").Return(nil)
-		// once for ingress update, once for service update
-		updater.On("Update", test.entries).Return(nil).Times(2)
+var expectGetIngresses = func(client *fake.FakeClient, ingresses []*v1beta1.Ingress) {
+	client.On("GetIngresses", &k8s.NamespaceSelector{LabelName: "team", LabelValue: "theteam"}).Return(ingresses, nil)
+}
```
howardburgess: Above are two expectations on the client calls, based on whether we're calling `GetAllIngresses` (if there was no namespace selector argument specified) or `GetIngresses(selector)` (if a selector was passed)

Since we're asserting on the `client` calls now, in addition to the `updater` calls, we need to pass in the expectations required.

controller/controller_test.go
```diff
@@ -270,508 +277,815 @@ func TestUnhealthyIfUpdaterFails(t *testing.T) {
...
+	ingressCh <- struct{}{}
+	serviceCh <- struct{}{}
+	namespaceCh <- struct{}{}
+	time.Sleep(smallWaitTime)
+
+	//then
+	assert.NoError(controller.Stop())
+	time.Sleep(smallWaitTime)
+	updater.AssertExpectations(t)
+	client.AssertExpectations(t)
```
howardburgess: Assert on the `client` calls, to make sure it's receiving the correct `GetAllIngresses`/`GetIngresses(namespaceSelector)` call.

controller/controller.go
```diff
@@ -184,23 +200,28 @@ func (c *controller) updateIngresses() (err error) {
...
+							ServicePort:    int32(path.Backend.ServicePort.IntValue()),
+							Allow:          c.defaultAllow,
+							StripPaths:     c.defaultStripPath,
+							ExactPath:      c.defaultExactPath, BackendTimeoutSeconds: c.defaultBackendTimeout,
 							BackendMaxConnections: c.defaultBackendMaxConnections,
 							ProxyBufferSize:       c.defaultProxyBufferSize,
 							ProxyBufferBlocks:     c.defaultProxyBufferBlocks,
 							CreationTimestamp:     ingress.CreationTimestamp.Time,
 							Ingress:               ingress,
+							IngressName:           ingress.Annotations[ingressNameAnnotation],
```
totahuanocotl: this is also super confusing. 
```suggestion
							IngressControllerName:           ingress.Annotations[ingressControllerNameAnnotation],
```

controller/controller.go
```diff
@@ -302,6 +323,17 @@ func (c *controller) updateIngresses() (err error) {
 	return nil
 }
 
+func (c *controller) ingressNameSupported(ingress *v1beta1.Ingress) bool {
+
+	isValid := false
+
+	if ingressName, ok := ingress.Annotations[ingressNameAnnotation]; ok {
```
totahuanocotl:
```suggestion
	if ingressControllerName, ok := ingress.Annotations[ingressControllerNameAnnotation]; ok {
```

controller/ingress_entry.go
```diff
@@ -13,6 +13,8 @@ type IngressEntries []IngressEntry
 
 // IngressEntry describes the ingress for a single host, path, and service.
 type IngressEntry struct {
+	// Name of the ingress to use
```
totahuanocotl:
```suggestion
	// Name of the ingress controller to use
```

controller/ingress_entry.go
```diff
@@ -13,6 +13,8 @@ type IngressEntries []IngressEntry
...
+	IngressName string
```
totahuanocotl:
```suggestion
	IngressControllerName string
```

elb/elb.go
```diff
@@ -141,8 +150,13 @@ func (e *elb) attachToFrontEnds() error {
 	return nil
 }
 
-// FindFrontEndElbs finds all elbs tagged with 'sky.uk/KubernetesClusterFrontend=<labelValue>'
-func FindFrontEndElbs(awsElb ELB, labelValue string) (map[string]LoadBalancerDetails, error) {
+// FindFrontEndElbs supports finding elbs without Ingress Name for backwards compatibility
+func FindFrontEndElbs(awsElb ELB, frontendTagValue string) (map[string]LoadBalancerDetails, error) {
+	return FindFrontEndElbsWithIngressName(awsElb, frontendTagValue, "")
+}
```
totahuanocotl: I thought the README specified this was a breaking change and that ingresses without tags would not attach to anything, but it seems we have backwards compatibility.

elb/status/status.go
```diff
@@ -18,9 +18,10 @@ import (
 
 // Config for creating a new ELB status updater.
 type Config struct {
-	Region           string
-	LabelValue       string
-	KubernetesClient k8s.Client
+	Region              string
+	FrontendTagValue    string
+	IngressNameTagValue string
```
totahuanocotl:
```suggestion
	IngressControllerNameTagValue string
```

elb/status/status.go
```diff
@@ -31,23 +32,25 @@ func New(conf Config) (controller.Updater, error) {
...
+	frontendTagValue    string
+	ingressNameTagValue string
+	loadBalancers       map[string]v1.LoadBalancerStatus
+	kubernetesClient    k8s.Client
 }
 
 // Start discovers the elbs and generates loadBalancer statuses.
 func (s *status) Start() error {
-	clusterFrontEnds, err := elb.FindFrontEndElbs(s.awsElb, s.labelValue)
+	clusterFrontEnds, err := elb.FindFrontEndElbsWithIngressName(s.awsElb, s.frontendTagValue, s.ingressNameTagValue)
```
totahuanocotl:
```suggestion
	clusterFrontEnds, err := elb.FindFrontEndElbsWithIngressControllerName(s.awsElb, s.frontendTagValue, s.ingressNameTagValue)
```

cmd/feed-ingress/main.go
```diff
@@ -167,6 +174,11 @@ func init() {
 			" '/myhost/myapp/health/'. Can be overridden with the sky.uk/exact-path annotation per ingress")
 	flag.IntVar(&healthPort, "health-port", defaultHealthPort,
 		"Port for checking the health of the ingress controller on /health. Also provides /debug/pprof.")
+	flag.StringVar(&ingressName, "ingress-name", defaultIngressName,
+		"Which ingress this instance is for.  This is used to filter ingresses so that only those with this"+
+			"ingress name will be updated.")
+	flag.StringVar(&namespaceSelector, namespaceSelectorFlag, defaultNamespaceSelector,
+		"Only consider ingresses within namespaces having labels matching this selector (e.g. app=loadtest).")
```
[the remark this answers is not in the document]
howardburgess: I'm not sure. Since this is an open-source project I think we should stick with the Kubernetes notion of labels/annotations, which says:

> Labels can be used to select objects and to find collections of objects that satisfy certain conditions. In contrast, annotations are not used to identify and select objects.

https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#attaching-metadata-to-objects

k8s/client.go
```diff
@@ -87,7 +102,48 @@ func (c *client) GetIngresses() ([]*v1beta1.Ingress, error) {
...
+
+	filteredIngresses := []*v1beta1.Ingress{}
+	for _, ingress := range allIngresses {
+		if ingressInSupportedNamespace(ingress, selector, c.namespaceStore.List()) {
+			filteredIngresses = append(filteredIngresses, ingress)
+		}
+	}
+
+	return filteredIngresses, nil
+}
```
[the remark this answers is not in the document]
howardburgess: I've made `GetAllIngresses` call `GetIngresses`, passing an empty (`nil`) namespace selector. This tidies things up and reduces the repetition. Lazy though?

k8s/client.go
```diff
@@ -87,7 +102,48 @@ func (c *client) GetIngresses() ([]*v1beta1.Ingress, error) {
...
+	}
+
+	allIngresses, err := c.GetAllIngresses()
+	if err != nil {
+		return nil, err
+	}
+
+	filteredIngresses := []*v1beta1.Ingress{}
+	for _, ingress := range allIngresses {
+		if ingressInSupportedNamespace(ingress, selector, c.namespaceStore.List()) {
```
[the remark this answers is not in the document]
howardburgess: It is the whole list, but my thinking about moving it to `controller.updateIngresses()` is that `updateIngresses()` is called on a watcher update for any of the three resource types being watched (ingresses, namespaces and services). Hence any filtering we do there would still happen at every single update.

We could have a different callback method for the namespace watcher, which performs this filtering though. Is it worth the extra complication though, since even on a namespace update we need to update Feed's IngressEntries.

README.md
```diff
@@ -17,9 +17,37 @@ can be applied to a cluster.
 
 ## Requirements
 
-* An internal and internet-facing ELB has been created and can reach your kubernetes cluster. The ELBs should be tagged with `sky.uk/KubernetesClusterFrontend=<name>` which is used by feed to discover them.
+* An internal and internet-facing ELB has been created and can reach your kubernetes cluster.
+The ELBs should be tagged with `sky.uk/KubernetesClusterFrontend=<name>` which is used by feed to discover them.
+If you are using v2 of `feed-ingress` the ELBs should also be tagged with `sky.uk/KubernetesClusterIngressName=<name>`.
```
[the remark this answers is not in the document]
howardburgess: This (and other comments) really does help in understanding this feature, I agree.

cmd/feed-ingress/main.go
```diff
@@ -134,6 +137,10 @@ const (
 	defaultClientHeaderBufferSize            = 16
 	defaultClientBodyBufferSize              = 16
 	defaultLargeClientHeaderBufferBlocks     = 4
+	defaultIngressName                       = ""
+	defaultNamespaceSelector                 = ""
+
+	namespaceSelectorFlag = "ingress-namespace-selector"
```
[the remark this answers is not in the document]
howardburgess: You mean to clarify what `defaultNamespaceSelector = ""` means? It's that there is none, hence all namespaces would be considered. I can clarify as a comment.

cmd/feed-ingress/main.go
```diff
@@ -167,6 +174,11 @@ func init() {
 			" '/myhost/myapp/health/'. Can be overridden with the sky.uk/exact-path annotation per ingress")
 	flag.IntVar(&healthPort, "health-port", defaultHealthPort,
 		"Port for checking the health of the ingress controller on /health. Also provides /debug/pprof.")
+	flag.StringVar(&ingressName, "ingress-name", defaultIngressName,
+		"Which ingress this instance is for.  This is used to filter ingresses so that only those with this"+
```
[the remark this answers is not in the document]
howardburgess: This was left over from the original PR from February, when extending this change to feed-dns was being considered. I have reverted the comment.

cmd/feed-ingress/main.go
```diff
@@ -238,8 +250,9 @@ func init() {
 	// elb/alb flags
 	flag.StringVar(&region, "region", defaultRegion,
 		"AWS region for frontend attachment.")
-	flag.StringVar(&elbLabelValue, "elb-label-value", defaultElbLabelValue,
-		"Attach to ELBs tagged with "+elb.ElbTag+"=value. Leave empty to not attach.")
+	flag.StringVar(&elbFrontendTagValue, "elb-frontend-tag-value", defaultElbFrontendTagValue,
+		"Alias to ELBs tagged with "+elb.ElbTag+"=value. Route53 entries will be created to these,"+
+			"depending on the scheme and ingress name.")
```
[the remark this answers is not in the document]
howardburgess: This was left over from the original PR from February, when extending this change to feed-dns was being considered. I have reverted the comment.

elb/elb.go
howardburgess: This method is still required because feed-dns isn't being updated for multiple ingress controllers, hence it needs the full list.

cmd/feed-ingress/main.go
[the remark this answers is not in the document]
howardburgess: You mean to clarify what defaultNamespaceSelector = "" means? It's that there is none, hence all namespaces would be considered, regardless of their labels.

k8s/client.go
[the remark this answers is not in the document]
totahuanocotl: We should still reduce the complexity of the current fitlering implementation for the worst case scenario from:
`All ingresses* All namespaces`
to
`All namespaces + All ingresses*Selected namespaces`

cmd/feed-ingress/main.go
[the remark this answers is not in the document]
totahuanocotl: jejeje. no sorry, just being picky. If you add a variable namespaceSelectorFlag constant, do them all :joy: (jk)

jsravn:
What does "multiple ingresses per environment." mean? Can you give an example? Thanks :).

howardburgess:
@jsravn the title should more accurately be "Multiple feed-ingress controllers per environment".

Currently only one feed-ingress can attach to ELBs and process ingresses for a particular environment.

In a multi-tenanted cluster, certain teams may have a requirement to segregate their traffic from others' using their own ELBs and feed-ingress instances. In this case, after this change, the following would allow traffic to be processed by an independent feed-ingress instance:

* Add an ELB with the new additional tag `sky.uk/KubernetesClusterIngressControllerName=xxx`
* Start a second feed-ingress instance with the `-ingress-controller-name=xxx` flag. It will attach to the matching ELB
* Annotate the ingress resource with `sky.uk/ingress-controller-name=xxx` to indicate that the matching feed-ingress should adopt this ingress resource and send traffic to its service

In addition to teams with this particular requirement, additional feed-ingress controllers can be used for load testing, for example, without disturbing existing traffic.

feed-ingress/cmd/alb.go
```diff
@@ -20,6 +20,15 @@ Specifically, they don't respect the deregistration delay. As a result, we don't
...
 	rootCmd.AddCommand(albCmd)
+
+	elbCmd.Flags().StringVar(&region, "region", defaultRegion,
+		"AWS region for frontend attachment.")
+	elbCmd.Flags().StringSliceVar(&targetGroupNames, "alb-target-group-names", []string{},
+		"Names of ALB target groups to attach to, separated by commas.")
+	elbCmd.Flags().DurationVar(&targetGroupDeregistrationDelay, "alb-target-group-deregistration-delay",
+		defaultTargetGroupDeregistrationDelay,
+		"Delay to wait for feed-ingress to deregister from the ALB target group on shutdown. Should match"+
+			" the target group setting in AWS.")
```
totahuanocotl: they are still being added to the elbCmd :(

cmd/feed-ingress/main.go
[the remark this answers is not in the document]
totahuanocotl: Perhaps it would even be beter to use the namespace.labels["team"]?
`group: DL-CIS SE Platform Technology - Pom` vs `team: cats`

feed-ingress/cmd/alb.go
howardburgess: D'oh!

howardburgess:
@totahuanocotl I have added the `--include-unnamed-ingresses` flag as discussed in yesterday's demo and with Joe afterwards. It's introduced as deprecated since it's for migration only.

howardburgess:
Travis is complaining about formatting. Goodness knows why... locally works fine:

```
$ git status
On branch named-ingresses
Your branch is up-to-date with 'origin/named-ingresses'.

nothing to commit, working tree clean

$ git rev-parse HEAD
fdb06aa7034c521690c8e56f573f66b545455a82

$ go version
go version go1.11.12 linux/amd64

$ make checkformat
== check formatting

$ echo $?
0
```

elb/elb.go
```diff
@@ -45,7 +45,7 @@ func New(region string, frontendTagValue string, ingressControllerNameTagValue s
 		metadata:                      ec2metadata.New(session),
 		awsElb:                        aws_elb.New(session),
 		frontendTagValue:              frontendTagValue,
-		ingressControllerNameTagValue: ingressControllerNameTagValue,
+		ingressControllerNameTagValue: ingressClassTagValue,
```
totahuanocotl: elb.ingressControllerNameTagValue -> el.ingressClassTagValue

Issue #192, opened 2018-12-14
Feed returns incorrect 301 Location header


shin-nien:
For any ingress that uses a non `/` path e.g. `/myapp`, trying to hit `http://<host>/myapp` results in a 301 with an incorrect `Location` header. Hitting `/myapp/` hits the backend as expected.

```
HTTP/1.1 301 Moved Permanently
Content-Type: text/html
Date: Fri, 14 Dec 2018 12:11:47 GMT
Location: http://shin-test-ingress.sandpfm.cosmic.sky:8080/myapp/
Server: nginx
Content-Length: 178
Connection: keep-alive
```

The `8080` is the port nginx is configured to listen to rather than the listening port of the ELB.

Arguably, hitting `/myapp` shouldn't be issuing a redirect at all; it's the implementation of https://github.com/sky-uk/feed/blob/319130ef7cd59301b61651174c6107e0556a3073/nginx/nginx.tmpl#L161 that appends a trailing `/` presumably to enforce exact matching that users of `feed` then had to adhere to.

Historically, most endpoints in `feed` have been APIs and this was only noticed when someone tried to use a path with a website and a browser although in practise, we tend to just use `/` as the path if we're fronting Grafana, Jenkins etc.

jsravn:
Is it possible to override 8080 w/ the correct port?

shin-nien:
That could work. The commandline args would just need to change to match the ELB ports and to check if the ingress is wanting https or not.

https://github.com/sky-uk/feed/blob/319130ef7cd59301b61651174c6107e0556a3073/nginx/nginx.tmpl#L150

jsravn:
A little harder than that probably, since you won't be able to bind to 80/443 by default without setting capabilities.

PR #197, opened 2019-03-14
Add support for exact paths when specified as locations

lw346:
Goes some way to resolving https://github.com/sky-uk/feed/issues/192.

Right now, if an ingress specifies a path of e.g. `/my/test/path`, this will be rendered in the location block as `location /my/test/path/`.  If a browser then hits `/my/test/path` it will be redirected to `/my/test/path/` with a 301 permanent redirect.

If `/my/test/path` is not a directory, but is instead an absolute path to a resource, then redirecting to `/my/test/path/` simply doesn't make sense.  This PR introduces the concept of "exact paths" by allowing ingresses to specify if all of their paths are absolute paths to resources.  If so, we render the path as `location = <raw path value>`.


cmd/feed-ingress/main.go
```diff
@@ -158,6 +159,11 @@ func init() {
 			"limitations. URL encoded characters will not work correctly in some cases, and backend services will "+
 			"need to take care to properly construct URLs, such as by using the 'X-Original-URI' header."+
 			"Can be overridden with the sky.uk/strip-path annotation per ingress")
+	flag.BoolVar(&controllerConfig.DefaultExactPath, "ingress-exact-path", defaultIngressExactPath,
+		"Whether to consider the ingress path to be an exact match rather than as a prefix. For example, "+
+			"if enabled 'myhost/myapp/health' would be match as 'myhost/myapp/health' but not 'myhost/myapp/health/x'."+
```
howardburgess: > if enabled 'myhost/myapp/health' would ~be~ match ~as~ 'myhost/myapp/health' but not 'myhost/myapp/health/x'.

alangibson01:
Sorry, @howardburgess, one more time please - I messed up the version number in the changelog.

PR #205, opened 2019-06-03
Skip ingress when http and/or path are not defined

rewiko:
Ingress can be created without Http and path, feed should skipped ingress definition withouth host definition.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -172,106 +172,112 @@ func (c *controller) updateIngresses() error {
 	var entries []IngressEntry
 	for _, ingress := range ingresses {
 		for _, rule := range ingress.Spec.Rules {
-			for _, path := range rule.HTTP.Paths {
 
-				serviceName := serviceName{namespace: ingress.Namespace, name: path.Backend.ServiceName}
+			if rule.HTTP != nil {
+				for _, path := range rule.HTTP.Paths {
 
-				if address := serviceMap[serviceName]; address != "" {
-					entry := IngressEntry{
-						Namespace:             ingress.Namespace,
-						Name:                  ingress.Name,
-						Host:                  rule.Host,
-						Path:                  path.Path,
-						ServiceAddress:        address,
-						ServicePort:           int32(path.Backend.ServicePort.IntValue()),
-						Allow:                 c.defaultAllow,
-						StripPaths:            c.defaultStripPath,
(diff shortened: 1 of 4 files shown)


controller/controller.go
```diff
@@ -172,106 +172,112 @@ func (c *controller) updateIngresses() error {
 	var entries []IngressEntry
 	for _, ingress := range ingresses {
```
domleb: Should we also wrap this processing in a try catch?

controller/controller_test.go
```diff
@@ -845,22 +881,52 @@ func createIngressesFixture(host string, serviceName string, servicePort int, in
 		}
 	}
 
-	return []*v1beta1.Ingress{
-		{
-			ObjectMeta: v1.ObjectMeta{
-				Name:        ingressName,
-				Namespace:   ingressNamespace,
-				Annotations: annotations,
+	if path == "" && host != "" {
```
totahuanocotl: you could call:
`ingress:= createIngressWithoutRules()`
and then just populate the Spec fields as needed to avoid some of the repetition

rewiko: made the change

controller/controller.go
rewiko: try/catch doesn't exist in golang, not sure how to improve the function

controller/controller_test.go
```diff
@@ -845,23 +881,27 @@ func createIngressesFixture(host string, serviceName string, servicePort int, in
...
+			Host:             host,
+			IngressRuleValue: v1beta1.IngressRuleValue{},
+		})
+	} else if path != "" && host != "" {
+		ingressRules = append(ingressRules, v1beta1.IngressRule{
+			Host: host,
+			IngressRuleValue: v1beta1.IngressRuleValue{HTTP: &v1beta1.HTTPIngressRuleValue{
+				Paths: paths,
+			}},
+		})
```
totahuanocotl: I think you can still simplify this a bit more:
```go
if host != "" {
   ingressRuleValue := v1beta1.IngressRuleValue{}
   ingressRule := v1beta1.IngressRule{
      Host:             host,
      IngressRuleValue: ingressRuleValue,
   }
   if path != "" {
      ingressRuleValue.HTTP =  &v1beta1.HTTPIngressRuleValue{ Paths: paths }
  }
  append(ingressRules, ingressRule)
}
```

PR #209, opened 2019-09-20
Add support for NLBs

howardburgess:
NLBs are supported by the AWS `elbv2` API/library, hence we are using both libraries to support ELBs and NLBs.

Also, for NLBs we attach to their TargetGroups rather than to the load balancer itself.

Grab me if you would like a walkthrough.

Required by https://github.com/sky-uk/core-infrastructure/issues/3906


feed-ingress/cmd/elb.go
```diff
@@ -36,11 +38,40 @@ var elbCmd = &cobra.Command{
...
+		t.value = defaultLoadBalancerType
+	}
+	return nil
+}
+
+func (t *loadBalancerTypeValue) Get() string {
+	if t.value == elb.Classic {
+		return "classic"
+	} else if t.value == elb.Standard {
+		return "standard"
```
domleb: I'm just thinking about the terminology here.  I don't think AWS have a 'standard' load balancer, just 'NLB' and 'ALB'.  So should this be named NLB as we also have an ALB package?  This also makes me wonder if we should have an NLB package instead of adding NLBs to the ELB package... might be easier to discuss in person / call.

controller/controller.go
```diff
@@ -301,12 +301,15 @@ func (c *controller) updateIngresses() (err error) {
 		}
 	}
 
-	log.Infof("Updating with %d entries", len(entries))
+	log.Infof("Updating with %d entries from %d total ingresses (skipped %d)", len(entries), len(ingresses), len(skipped))
 	if len(skipped) > 0 {
-		log.Infof("Skipped %d invalid: %s", len(skipped), strings.Join(skipped, ", "))
+		for _, msg := range skipped {
```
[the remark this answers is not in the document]
howardburgess: It could, yes. I've changed it to debug instead, so it's there for investigations.

feed-ingress/cmd/elb.go
howardburgess: You're right - and I like the idea of having an `nlb` subcommand - then we've got one subcommand for each of the AWS  load balancer types (Classic aka ELB, Network and Application)

PR #213, opened 2019-11-25
Handle receiving 0 ingresses and prevent http 404s

domleb:
Don't start Nginx or reload the config if Update is called with 0
ingresses.  Instead, return an error and let the controller handle it.
This fixes an issue where the k8s client returns 0 ingresses during
startup and Nginx then returns http 404s for all ingresses.

Fix for the sorting of ingresses when filtering for unique entries.
This fixes an issue with flapping ingresses when the same path is
defined for different services under the same host. This flapping
was causing Nginx to reload frequently which is disruptive.  We now
include service and port in the sorting.

Add a metric for Nginx config reloads.

Required by https://github.com/sky-uk/core-infrastructure/issues/5657

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -177,23 +177,40 @@ func (c *controller) updateIngresses() (err error) {
 		}
 	}()
 
+	// Get ingresses
 	var ingresses []*v1beta1.Ingress
+
 	if c.namespaceSelector == nil {
 		ingresses, err = c.client.GetAllIngresses()
 	} else {
 		ingresses, err = c.client.GetIngresses(c.namespaceSelector)
 	}
+
 	log.Infof("Found %d ingresses", len(ingresses))
+
 	if err != nil {
 		return err
 	}
+
+	if len(ingresses) == 0 {
+		return errors.New("found 0 ingresses")
+	}
+
+	// Get services
 	services, err := c.client.GetServices()
+
 	if err != nil {
 		return err
 	}
 
-	serviceMap := serviceNamesToClusterIPs(services)
+	log.Infof("Found %d services", len(services))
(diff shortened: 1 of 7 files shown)


controller/controller.go
```diff
@@ -177,23 +177,40 @@ func (c *controller) updateIngresses() (err error) {
...
+	// Get ingresses
 	var ingresses []*v1beta1.Ingress
+
 	if c.namespaceSelector == nil {
 		ingresses, err = c.client.GetAllIngresses()
 	} else {
 		ingresses, err = c.client.GetIngresses(c.namespaceSelector)
 	}
+
 	log.Infof("Found %d ingresses", len(ingresses))
```
mgreyo: Can we perhaps move this further down, so once it's checked that the length of `ingresses` and `services` is > 0, it will log as one line? e.g. `log.Infof("Found %d Ingresses and %d services", len(ingresses), len(services))`

At the moment we log what it's found, then we log the reason if it errors (the same message). If we keep the individual logs, can we change them to `log.Debugf()` just to reduce the noise? 
```
Found 0 ingresses
Unable to update ingresses: found 0 ingresses
```

```suggestion
	log.Debugf("Found %d ingresses", len(ingresses))

	if err != nil {
		return err
	}

	if len(ingresses) == 0 {
		return errors.New("found 0 ingresses")
	}

	// Get services
	services, err := c.client.GetServices()

	if err != nil {
		return err
	}

	log.Debugf("Found %d services", len(services))

	if len(services) == 0 {
		return errors.New("found 0 services")
	}

	log.Infof("Found %d Ingresses and %d services", len(ingresses), len(services))
```

PR #216, opened 2020-01-23
Remove http binding by default

rewiko:
Remove http binding by default, feed should be able to run with https only to avoid using an extra port.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# v3.0.0
+* Breaking change 
+  By default feed will not bind http port on 8080 if `--ingress-port=XXXX` is not provided   
+  * Remove http binding by default (#216)
+ 
+# v2.3.0
+* Support IP and Instance TargetGroup attachments (#214)
+
 # v2.2.3
 * [BUGFIX] Handle receiving 0 ingresses from the k8s client by not reloading the Nginx config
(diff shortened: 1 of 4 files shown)


feed-ingress/cmd/common_test.go
```diff
@@ -0,0 +1,42 @@
...
+		createPortsConfig(unset, unset)
+		return
+	}
+	cmd := exec.Command(os.Args[0], "-test.run=TestCreatePortsConfigWithoutPorts")
+	cmd.Env = append(os.Environ(), "BE_CRASHER=1")
+	err := cmd.Run()
+	if e, ok := err.(*exec.ExitError); ok && !e.Success() {
+		return
+	}
+	t.Fatalf("process ran with err %v, want exit status 1", err)
```
domleb: just a minor thing but we could say 'process exited with %v, want exit status 1", err' as I think this could be exit status 0

rewiko:
I will need to update the changelog.

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# v3.0.0
+* Breaking change 
+  By default feed will not bind http port on 8080 if `--ingress-port=XXXX` is not provided
```
howardburgess: Just to avoid the double-negative :)

```suggestion
  Feed will not bind port http 8080 by default, unless `--ingress-port=XXXX` is provided   
```

PR #218, opened 2020-03-13
Wait for informer cache to be fully populated before interrogating it

sebbonnet:
This fixes an issue where feed-ingress occassionally returns 404 for requests to a subset of the virtual hosts it manages.
This would happen on feed-ingress startup and would continue until the next nginx reload happens.
The logs would show a sequence of events similar to this where the first watcher update would return a partial list of ingresses:
```
time="2020-02-19T17:55:02Z" level=info msg="Received update on watcher" source="controller/controller.go:160"
time="2020-02-19T17:55:02Z" level=info msg="Found 63 ingresses and 527 services" source="controller/controller.go:212"
time="2020-02-19T17:55:02Z" level=info msg="Updating with 94 entries from 63 total ingresses (skipped 0)" source="controller/controller.go:335"
time="2020-02-19T17:55:02Z" level=info msg="Updating nginx config: --- [...]
time="2020-02-19T17:55:02Z" level=info msg="Starting nginx for the first time" source="nginx/nginx.go:211"
[...]
time="2020-02-19T17:55:03Z" level=info msg="Received update on watcher" source="controller/controller.go:160"
time="2020-02-19T17:55:03Z" level=info msg="Found 620 ingresses and 527 services" source="controller/controller.go:212"
time="2020-02-19T17:55:03Z" level=info msg="Updating with 839 entries from 620 total ingresses (skipped 18)" source="controller/controller.go:335"
[...]
time="2020-02-19T18:00:02Z" level=info msg="Signalling Nginx to reload configuration" source="nginx/nginx.go:93"
```

When creating an informer the store cache is populated asynchronously. Interrogating the cache before it's fully populated may return a partial update or even 0 entries - this is the reason we've seen updates with 0 ingresses in the past. 

This issue was fixed by waiting for the store cache to be fully populated before interrogating it.
There should now not be any cases for updates to return 0 ingresses and so we could remove the extra safety checks introduced by https://github.com/sky-uk/feed/pull/213. In fact in my testing I've seen no such occurrences in the logs anymore.


k8s/store.go
```diff
@@ -0,0 +1,156 @@
...
+		clientset:           clientset,
+		stopCh:              stopCh,
+		resyncPeriod:        resyncPeriod,
+		informerFactory:     &cacheInformerFactory{clientset: clientset},
+		eventHandlerFactory: &bufferedEventHandlerFactory{},
+	}
+}
+
+//  GetOrCreateNamespaceSource creates a namespace informer and registers and event handler to watch for changes
+func (s *lazyLoadedStore) GetOrCreateNamespaceSource() (*WatchedResource, error) {
```
totahuanocotl: I think this method would need to be guarded by locks.

k8s/store.go
```diff
@@ -0,0 +1,156 @@
...
+	informerFactory          informerFactory
+	eventHandlerFactory      eventHandlerFactory
+	namespaceWatchedResource *WatchedResource
+	ingressWatchedResource   *WatchedResource
+	serviceWatchedResource   *WatchedResource
+}
+
+// WatchedResource exposes a store for the resource to allow interrogating the current state via get/list operations
+// and an watcher so clients can subscribe to update notifications
+type WatchedResource struct {
```
alangibson01: Just a minor quibble with the naming: when I first read `WatchedResource` elsewhere in the code, I thought it was referring to a specific individual named object, e.g. a single ingress.

k8s/store.go
```diff
@@ -0,0 +1,156 @@
...
+
+	if !cache.WaitForCacheSync(s.stopCh, controller.HasSynced) {
+		return &WatchedResource{}, errors.New("error while waiting for namespace cache to populate")
+	}
+
+	log.Debug("Namespace cache has been fully populated")
+	return &WatchedResource{
+		store:   store,
+		watcher: namespaceWatcher,
+	}, nil
```
alangibson01: I think there's scope for refactoring this out into code shared with the subsequent two methods. It all looks common apart from the `informerFactory` call on line 82, but that function could be passed as an argument into an extracted common function.

k8s/event_handler.go
```diff
@@ -0,0 +1,48 @@
...
+	log "github.com/sirupsen/logrus"
+	"k8s.io/client-go/tools/cache"
+)
+
+type handlerWatcher struct {
+	*bufferedWatcher
+}
+
+// Implement cache.ResourceEventHandler
+var _ cache.ResourceEventHandler = &handlerWatcher{}
```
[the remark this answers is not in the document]
sebbonnet: at first it is :) but it's quite convenient to show when the implementation is no longer compliant

k8s/store.go
sebbonnet: good point

k8s/store.go
sebbonnet: sounds good

k8s/store.go
sebbonnet: I'm open to suggestion :)

Makefile
```diff
@@ -58,7 +58,8 @@ test : build
 # Docker build
 git_rev := $(shell git rev-parse --short HEAD)
 git_tag := $(shell git tag --points-at=$(git_rev))
-image_prefix := skycirrus/feed
+REGISTRY ?= skycirrus
```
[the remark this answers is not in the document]
sebbonnet: correct

CHANGELOG.md
```diff
@@ -1,8 +1,14 @@
+# v3.1.1
+* [BUGFIX] Wait for informer cache to be fully populates before interrogating it (#218)
```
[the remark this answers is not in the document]
sebbonnet: I agree but the versioning used for 3.1.0 doesn't suggest it was a bug, but rather a feature, otherwise it should have been 3.0.1. So I simply added the missing changelog respecting the choice of version

k8s/store.go
alangibson01: `WatchedStore`?

alangibson01:
I'm happy with this now.

sebbonnet:
@alangibson01 I've made some more changes to tackle a bug we found with @totahuanocotl and to make sure go routines created for the controllers would not be left behind on shutdown: there is now a single stop channel that is passed down from the top level (main.go and common.go)

alangibson01:
@sebbonnet Looks fine to me now, sorry for not catching the channel thing and the other bits sooner. Any more changes on the way?

sebbonnet:
No more changes :)

alangibson01:
OK from me, approval required from @totahuanocotl in order to merge.

controller/controller.go
```diff
@@ -136,19 +136,32 @@ func (c *controller) Start() error {
...
+	}
+	namespaceWatcher, err := c.client.WatchNamespaces()
+	if err != nil {
+		return err
+	}
 	c.watcher = k8s.CombineWatchers(ingressWatcher, serviceWatcher, namespaceWatcher)
 	c.watcherDone.Add(1)
 	go c.handleUpdates()
+	return nil
 }
```
totahuanocotl: This is actually changing Feed's startup behaviour. 
Before, Feed would start up, even if there were sync errors on the k8s.client resources watchers. If the client remained `unsynced`, this would be caught on the `updateIngresses` cycle, and set feed in an unhealthy state.
 
Now, if for some reason Feed gets an error while starting to watch the resources, Feed will error and not start.
This could send feed into a crash loop state depending on the cause of the error. 

Although undocumented and untested, I don't think we should change that behaviour.

k8s/client.go
```diff
@@ -86,28 +71,37 @@ func New(kubeconfig string, resyncPeriod time.Duration) (Client, error) {
...
+	}
+	return c, nil
 }
 
 func (c *client) GetAllIngresses() ([]*v1beta1.Ingress, error) {
 	return c.GetIngresses(nil)
 }
 
 func (c *client) GetIngresses(selector *NamespaceSelector) ([]*v1beta1.Ingress, error) {
-	if !c.namespaceController.HasSynced() {
```
totahuanocotl: To avoid the ingress partial updates, we should include the check for the `ingressController` here as well.

totahuanocotl:
> No more changes :)

Oops, I did bring up more changes :(

sebbonnet:
@totahuanocotl I've reverted the behaviour around the `Watchers` to ensure feed can start even if communication with the apiserver is not possible. I initially wanted to both document and test this behaviour, but I think this would be best done in an e2e test. For now I've simply documented the behaviour in the `client.go` file to make it easier to remember. I've leveraged the work I had done to test the store and added tests for `client.go` in the process to test my changes. 

As we're now no longer waiting until the ingress store cache is fully populated, the control loop does find the ingress cache not completely synced, but this only shows at most a couple of times on startup so I think this is fine - see `Unable to update ingresses: ingresses haven't synced yet`
```
time="2020-03-17T18:02:57Z" level=info msg="ELB Front end region: eu-west-1, cluster: sandfun, expected frontends: 2, ingress controller: main" source="elb/elb.go:37"
time="2020-03-17T18:02:57Z" level=info msg="nginx version: nginx/1.15.7" source="runtime/asm_amd64.s:1357"
time="2020-03-17T18:02:57Z" level=info msg="Creating nginx.conf for the first time" source="nginx/nginx.go:334"
time="2020-03-17T18:02:57Z" level=info msg="Found frontend elb sandfun-frontend-internal" source="elb/elb.go:217"
time="2020-03-17T18:02:57Z" level=info msg="Found frontend elb sandfun-frontend-internet-facing" source="elb/elb.go:217"
time="2020-03-17T18:02:57Z" level=info msg="Controller started" source="cmd/common.go:54"
time="2020-03-17T18:02:57Z" level=info msg="Received update on watcher" source="controller/controller.go:160"
time="2020-03-17T18:02:57Z" level=error msg="Unable to update ingresses: ingresses haven't synced yet" source="controller/controller.go:163"
time="2020-03-17T18:02:57Z" level=info msg="Received update on watcher" source="controller/controller.go:160"
time="2020-03-17T18:02:57Z" level=error msg="Unable to update ingresses: ingresses haven't synced yet" source="controller/controller.go:163"
time="2020-03-17T18:02:57Z" level=info msg="Received update on watcher" source="controller/controller.go:160"
time="2020-03-17T18:02:57Z" level=info msg="Found 1007 ingresses and 9 services" source="controller/controller.go:212"
time="2020-03-17T18:02:57Z" level=info msg="Updating with 1007 entries from 1007 total ingresses (skipped 0)" source="controller/controller.go:335"
time="2020-03-17T18:02:57Z" level=info msg="Updating nginx config: ---
```
 
I kept the changes related to the single stop channel to make sure go routines created for the controllers would not be left behind on shutdown.

I also considered removing the `ingress == 0` check introduced by #213, but I realised that it's still useful to have to cover cases where there are no ingresses for a given namespace for instance.

controller/controller.go
sebbonnet: I agree. The client api was actually designed to make it impossible for Watcher(s) to return errors, so feed doesn't have to rely on the apiserver to be up and running to start. In fact it is expected that errors are handled during the control loop, so get calls that fail due to cache not being fully synced or other can be retried.

k8s/client.go
sebbonnet: yeah - this is where it used to be, but was unintentionally removed in a previous PR...

k8s/client.go
```diff
@@ -187,8 +187,6 @@ func (c *client) createIngressSource() {
 }
 
 func (c *client) GetServices() ([]*v1.Service, error) {
-	c.createServiceSource()
```
sebbonnet: I can't think of any reasons why this was needed, given that similarly to ingresses we first watch services before we attempt to retrieve them.

controller/controller_test.go
```diff
@@ -262,9 +262,9 @@ func TestUnhealthyIfUpdaterFails(t *testing.T) {
 
 	client.On("GetAllIngresses").Return(createDefaultIngresses(), nil)
 	client.On("GetServices").Return(createDefaultServices(), nil)
-	client.On("WatchIngresses").Return(ingressWatcher)
-	client.On("WatchServices").Return(serviceWatcher)
-	client.On("WatchNamespaces").Return(namespaceWatcher)
+	client.On("WatchIngresses").Return(ingressWatcher, nil)
```
totahuanocotl: I think these need to be reverted as well. `WatchXxx()` no longer returns a tuple

sebbonnet: well spotted!
I've now reverted - mind reapproving?

totahuanocotl:
> I also considered removing the `ingress == 0` check introduced by #213, but I realised that it's still useful to have to cover cases where there are no ingresses for a given namespace for instance.

I think we can revisit some of the changes from #213 . Probably the one on the control loop is fine, but the one that prevents nginx to start is not.
Even if nginx starts, if the ingresses are 0 in the cluster, feed will never become ready, so traffic will not reach it. Removing this from nginx, will mean that feed will not be killed due to liveness failures if there are actually 0 ingresses.

sebbonnet:
> I think we can revisit some of the changes from #213 . Probably the one on the control loop is fine, but the one that prevents nginx to start is not.
Even if nginx starts, if the ingresses are 0 in the cluster, feed will never become ready, so traffic will not reach it. Removing this from nginx, will mean that feed will not be killed due to liveness failures if there are actually 0 ingresses.

Can we do this as a follow up?
I feel it's best to keep the scope of this PR to fixing the ingress partial update

totahuanocotl:
> Can we do this as a follow up?
> I feel it's best to keep the scope of this PR to fixing the ingress partial update

yeah, that's why I approved it.

alangibson01:
Nothing further from me.

PR #219, opened 2020-03-25
Account for concurrent status updates

alangibson01:
The update of an ingress status by a given feed pod F1 is a get-modify-set operation, where another feed instance F2 may potentially complete the set operation while F1 is still at the modify stage.

What this means is that when F1 comes to do its set operation, it will get back from the API server a "409 Conflict" response, indicating that someone else has modified the resource after F1 fetched it.

This change makes the update routine a little more resilient. If a feed pod receives a conflict response back when it attempts to update the status, it will perform a second get operation, and compare the status on the returned object with the status it wanted to set. If these are equal, it will not return any error.

In the case where the status on the returned object is not equal to the value it wanted to set, it will surface the error exactly as it does right now. Likewise, if the error is anything other than a conflict, it
will be returned exactly as it is at the moment.

### DIFF
--- a/k8s/client.go
+++ b/k8s/client.go
@@ -11,7 +11,9 @@ import (
 	"sync"
 	"time"
 
+	k8errors "k8s.io/apimachinery/pkg/api/errors"
 	clientV1Beta1 "k8s.io/client-go/kubernetes/typed/extensions/v1beta1"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


k8s/client.go
```diff
@@ -25,6 +31,16 @@ import (
...
 const bufferedWatcherDuration = time.Millisecond * 50
 
+var once sync.Once
+var conflictingIngressStatusUpdates prometheus.Counter
+
+func initMetrics() {
+	once.Do(func() {
+		conflictingIngressStatusUpdates = metrics.RegisterNewDefaultCounter("client", "conflicting_status_updates",
+			"Count of ingress status updates which resulted in a conflict, but which did not cause an error")
+	})
```
jsravn, line 41: Is it necessary to have a metric for this? I feel since this is just meta information anyways, this is not something we need to be too worried about.

k8s/client.go
```diff
@@ -242,16 +259,47 @@ func (c *client) createNamespaceSource() {
...
+	_, err := ingressClient.UpdateStatus(ingress)
+
+	switch {
+	case k8errors.IsConflict(err):
+		updatedIng, getErr := ingressClient.Get(ingress.Name, metav1.GetOptions{})
+		if getErr == nil && ingressStatusEqual(updatedIng.Status.LoadBalancer.Ingress, ingress.Status.LoadBalancer.Ingress) {
+			conflictingIngressStatusUpdates.Inc()
+			return nil
+		}
+		return err
```
jsravn, line 283: Isn't this sort of overkill? Conflict will happen if multiple feeds try to update the status, which we know will always happen at startup. Given it is expected, couldn't we just ignore the error if it is a conflict error? So we wouldn't need more than
```go
_, err := ingressClient.UpdateStatus(currentIng)
if !k8errors.IsConflict(err) {
  // ignore conflicts, as they are expected
  return err
}
return nil
```

alangibson01: That's assuming the conflict is for the reason we expect. Unlikely though it may be that we'll see a conflict for another reason, I don't think the additional logic adds that much in the way of complexity.

k8s/client.go
alangibson01: Yeah, I guess it's not that relevant.

k8s/client.go
jsravn: Okay, fair enough - would you mind adding a comment describing what it's checking? As I don't think it's super obvious why it returns nil in this case.

alangibson01:
@jsravn Comment added, ready for another review.

alangibson01:
As always, I forgot to update the changelog :-(

PR #220, opened 2020-03-27
Validate the sky.uk/allow annotation entries

alangibson01:
Each comma-separated entry should be either an IP address or a CIDR range. Ingresses where this isn't the case will be skipped, with an error message logged detailing the invalid entries.

### DIFF
--- a/controller/ingress_entry.go
+++ b/controller/ingress_entry.go
@@ -3,6 +3,8 @@ package controller
 import (
 	"errors"
 	"fmt"
+	"net"
+	"strings"
 	"time"
 
 	"k8s.io/api/extensions/v1beta1"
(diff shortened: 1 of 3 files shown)


controller/ingress_entry.go
```diff
@@ -64,6 +66,22 @@ func (e IngressEntry) validate() error {
...
+			if _, _, err := net.ParseCIDR(allowEntry); err != nil {
+				invalidAllowEntries = append(invalidAllowEntries, allowEntry)
+			}
+		}
+	}
+
+	if len(invalidAllowEntries) == 1 {
+		return fmt.Errorf("invalid entry in sky.uk/allow: %s", invalidAllowEntries[0])
+	} else if len(invalidAllowEntries) > 1 {
+		return fmt.Errorf("invalid entries in sky.uk/allow: %s", strings.Join(invalidAllowEntries, ","))
```
sebbonnet: I think we should indicate which host the entry in invalid for  as there could be a large number of ingresses.

controller/ingress_entry.go
```diff
@@ -64,6 +66,22 @@ func (e IngressEntry) validate() error {
...
+	for _, allowEntry := range e.Allow {
+		if net.ParseIP(allowEntry) == nil {
+			if _, _, err := net.ParseCIDR(allowEntry); err != nil {
+				invalidAllowEntries = append(invalidAllowEntries, allowEntry)
+			}
+		}
+	}
+
+	if len(invalidAllowEntries) == 1 {
+		return fmt.Errorf("invalid entry in sky.uk/allow: %s", invalidAllowEntries[0])
```
sebbonnet: Any reason to have this additional block other than to change the message from "entry" to "entries"? `strings.Join` won't print the comma if there is only one element

controller/ingress_entry.go
sebbonnet: besides in when the entry is empty, it would be quite hard to find which one is missing

sebbonnet:
You should also add a note in the CHANGELOG.md :)

controller/ingress_entry.go
alangibson01: Actually, I ought to add a test that covers the case of empty entries as well.

controller/ingress_entry.go
alangibson01: Yeah, just being picky grammatically, but there isn't much need for that. I'll simplify the code.

alangibson01:
Changes made as requested.

PR #221, opened 2020-03-31
Additional information in README.md

alangibson01:
Stating that it may take up to 2 minutes for Feed pods to register with an NLB.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,8 +153,10 @@ See the [example deployment for ELB](examples/feed-ingress-deployment-elb.yml)
 ### Network Load Balancers (NLBs)
 See the [example deployment for NLB](examples/feed-ingress-deployment-nlb.yml)
 
-Please note that new Feed pods will not start handling requests until they have been registered with the NLB. This can
-take up to 2 minutes on some occasions.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -153,6 +153,9 @@ See the [example deployment for ELB](examples/feed-ingress-deployment-elb.yml)
 ### Network Load Balancers (NLBs)
 See the [example deployment for NLB](examples/feed-ingress-deployment-nlb.yml)
 
+Please note that new Feed pods will not start handling requests until they have been registered with the NLB. This can
+take up to 2 minutes on some occasions.
```
totahuanocotl, line 157: On the risk of sounding a bit to strict, can we describe it in therms of the [target health status](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-health-checks.html#target-health-states)?
It seems the delay we see is due to the `Initial` state, which involves `registration` and `initial` health checks.
The NLB will only send request to the feed instance once it has reached the `healthy` status.

totahuanocotl:
travis... :man_facepalming:

PR #223, opened 2020-04-06
Trim whitespace from sky.uk/allow entries

alangibson01:
Ensures that additional whitespace around ingress entries does not stop them being recognised as valid addresses or ranges.


alangibson01:
@mgreyo Forgot to update the changelog, sorry!

PR #224, opened 2020-04-07
Replace fake_graceful_nginx.py

alangibson01:
I haven't done a full root-cause analysis, but it appears that the nginx tests which were flaky on Travis were due to the use of a Python based nginx fake implementation. I've replaced it with an equivalent Go implementation which doesn't cause the same flaky behaviour.


nginx/fake/fake_graceful_nginx.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+
+	case <-timer.C:
+		fmt.Println("Quit after 5 seconds of nada")
+		os.Exit(-1)
+	}
+}
+
+func startupMarkerFilename() string {
+	filename := strings.Split(os.Args[2], "/")
```
sebbonnet, line 58: would be nicer to pass `os.Args[2]` as an arg to the method

Makefile
```diff
@@ -52,6 +52,8 @@ lint :
 	done;
 
 test : build
+	@echo "== build fake nginx for tests"
+	@go build -o nginx/fake/fake_graceful_nginx nginx/fake/fake_graceful_nginx.go
```
sebbonnet, line 56: I think this should go in its own target, so that you can build the fake nginx binary and then run the test in your IDE.
the test target would then depend on that new target

alangibson01:
Good suggestions. Implemented as requested.

PR #227, opened 2020-06-11
Skip misconfigured vts entries

mgreyo:
Misconfigured ingress resources have the ability to create invalid
VTS entries. These are long-lived and are not rectified againt the
running nginx config.

Because of this, it is possible that the metrics exported by feed
can flip between the valid and invalid VTS data causing erroneous
metrics to be exported.

This adds an extra filter to ensure that metrics are not generated
for misconfigured ingress resources.

Required by https://github.com/sky-uk/core-infrastructure/issues/7091

### DIFF
--- a/nginx/nginx_test.go
+++ b/nginx/nginx_test.go
@@ -2044,6 +2044,12 @@ func TestUpdatesMetricsFromNginxStatusPage(t *testing.T) {
 	assertIngressRequestCounters(t,
 		"heapster.sandbox.cosmic.sky", "/",
 		2012.0, 1099.0, 0.0, 7.0, 0.0, 0.0, 0.0)
+	assertIngressRequestCounters(t,
+		"duplicate-path.sandbox.cosmic.sky", "/path/",
+		5000.0, 2000.0, 0.0, 5.0, 0.0, 0.0, 0.0)
+	assertIngressRequestCounters(t,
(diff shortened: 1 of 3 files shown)


nginx/nginx_test.go
```diff
@@ -2044,6 +2044,12 @@ func TestUpdatesMetricsFromNginxStatusPage(t *testing.T) {
 	assertIngressRequestCounters(t,
 		"heapster.sandbox.cosmic.sky", "/",
 		2012.0, 1099.0, 0.0, 7.0, 0.0, 0.0, 0.0)
+	assertIngressRequestCounters(t,
+		"duplicate-path.sandbox.cosmic.sky", "/path/",
+		5000.0, 2000.0, 0.0, 5.0, 0.0, 0.0, 0.0)
+	assertIngressRequestCounters(t,
+		"misconfigured-ingress.sandbox.cosmic.sky", "/bad/",
+		0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0)
```
totahuanocotl, line 2052: it is not really clear to me what is the difference between these two.
Is it to prove that a `bad path with a valid duplicate` will report the valid count, whereas a `bad path without a valid duplicate'  will report 0?

mgreyo: Correct. I've changed the names of these hosts and added comments that will hopefully express the tests intentions

PR #230, opened 2021-02-03
Cut down feed ingress vulnerabilities

sebbonnet:
To cut down on the attack surface
and so reduce the number of vulnerabilities.
The previous debian:strech-slim was showing a lot of vulnerabilities 
that have all gone away once switching to alpine (3.13).

Also used trivy to fail the build when either HIGH or CRITICAL
vulnerabilities are found.


sebbonnet:
trivy scan output is colourful :)
Screenshot from 2021-02-03 12-25-42

sebbonnet:
trivy scan output based on the new image:

Screenshot from 2021-02-03 12-25-42

sebbonnet:
This commit 908f7db showed how we used a kind to validate that feed would start and the vts plugin worked as expected. We've yet to check the opentracing plugin also works.
Using kind + mock aws service to run integration test would work pretty well (even in travis), but the test setup would need to be done in golang instead of bash to make it easier to maintain and read. This is something we could potentially do if people are interested / feel the value-add is worth if.
Until then we reverted that commit for now, so we have at least a trace of it.

sebbonnet:
This commit 908f7db showed how we used `kind` to validate that feed and the vts plugin would work once deployed. We've yet to check the opentracing plugin also works.
Using kind + mock aws service to run integration test would work pretty well (even in travis), but the test setup would need to be done in golang instead of bash to make it easier to maintain and read. This is something we could potentially do if people are interested / feel the value-add is worth it.
Until then we reverted that commit for now, so we have at least a trace of it.

sebbonnet:
> We've yet to check the opentracing plugin also works.

We've now tested opentracing too and found that it works fine. The path to the jaeger plugin was wrong, so we fixed it in b102ece.
We've also updated the integration tests we put together earlier with kind to also deploy an all-in-one jaeger using the jaeger-operator so we could make sure the opentracing plugin works and confirm we see jaeger traces when calling our test backend.
The integration tests is stored here: https://github.com/sky-uk/feed/compare/feed-integration-test-with-kind

rewiko (approved):
LGTM

PR #231, opened 2021-02-19
Specify the name of the header to use for deriving client ip

supreethrao:
- Specify the name of the header to use by the real ip module to derive the ip of the client to apply allow/deny rule.


rewiko (approved):
LGTM

rewiko (approved):

Issue #233, opened 2021-04-15
Feed instance abrupt termination when unable to de-register on a subset of attached NLB target groups


supreethrao:
Same Feed instance can attach to multiple target groups. During shutdown, it will then attempt to de-register itself from all of the target groups. In case of a failure to de-register from any of them, feed instance will not honour the shutdown or the drain interval and shutdown immediately hence causing an ungraceful shutdown for the valid registered target groups

https://github.com/sky-uk/feed/blob/master/nlb/nlb.go#L348-L359

PR #234, opened 2021-04-16
Fix bug during feed shutdown

supreethrao:
- Feed instance can attach to multiple target groups belonging to the load balancer having a tag matching the value in the flag `ingress-class`. The fix is for feed instance to wait for the drain duration even if the de-register call succeeds to at least one of those target groups.

closes https://github.com/sky-uk/feed/issues/233

### DIFF
--- a/nlb/nlb.go
+++ b/nlb/nlb.go
@@ -69,7 +69,7 @@ type nlb struct {
 	expectedNumber       int
 	instanceID           string
 	privateIPAddress     string
-	elbs                 map[string]LoadBalancerDetails
+	loadBalancers        map[string]LoadBalancerDetails
 	registeredFrontends  util.SafeInt
 	initialised          initialised
 	drainDelay           time.Duration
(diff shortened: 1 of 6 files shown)


nlb/nlb.go
```diff
@@ -342,54 +342,53 @@ func generateTargetDescriptionFromTargetType(targetType, instanceID, privateIP s
...
 func (e *nlb) Stop() error {
-	var failed = false
-	for _, elb := range e.elbs {
-		log.Infof("Deregistering instance %s with nlb %s", e.privateIPAddress, elb.Name)
-		err := deregisterFromLoadBalancer(e, elb)
-		if err != nil {
-			log.Warnf("unable to deregister instance %s (%s) with nlb %s: %v", e.instanceID, e.privateIPAddress, elb.Name, err)
-			failed = true
+	failedCount := 0
+	successCount := 0
```
jsravn, line 346: These could just be booleans I think.

supreethrao: Yeh sure. My initial idea was to use the number which failed for some elaborate logging but now that I did not do that, I can change those to booleans

rewiko (approved):

PR #236, opened 2021-09-18
Upgrade nginx and suppress server name response header

supreethrao:
- Upgrade nginx version to `1.21.3` from the current `1.15.7`
- Configure and add a dynamic module to suppress the server response header providing the name of the web server (nginx)

<details>
<summary>Manual verification of server response header suppression</summary>
<p>

```
Curl output before adding the directive

$ curl -XHEAD localhost:5090/hello -vv
< HTTP/1.1 404 Not Found
< Server: nginx
< Date: Sat, 18 Sep 2021 07:57:14 GMT
< Content-Type: text/html
< Content-Length: 146
< Connection: keep-alive
```
</p>

<summary>Curl output after adding the directive</summary>
<p>

```
$ curl -XHEAD localhost:5090/hello -vv
< HTTP/1.1 404 Not Found
< Date: Sat, 18 Sep 2021 07:56:38 GMT
< Content-Type: text/html
< Content-Length: 146
< Connection: keep-alive
```
</p>
</details>


chaosbox:
👍 
Do we have to copy the dynamic modules to the final docker image? `docker/ingress/Dockerfile`

aecay:
@chaosbox 

Looks to me like we build the .so into a dir under `/nginx` and then we [copy](https://github.com/sky-uk/feed/blob/a1bc2ff82be40fce24f7c9db0f330275132af13a/docker/ingress/Dockerfile#L48) that whole /nginx dir to the result.  So it should be ok already I think....

chaosbox:
Right I see it now.. thanks

chaosbox (approved):

PR #238, opened 2021-09-23
New external command for scenareos where LBs are attached externally not explicitly via feed

chbatey:
The motivation for this is to run feed in GKE.

Firstly we were going to follow a similar model to the NLBs are attach to pre-existing NLBs but in the end it was simpler to allow the GKE ingress controller to create and attach internal/external LBs to the feed pods and feed update each ingress to point to a static host. This requires a service/ingress for feed.

So looks like this:

image


howardburgess (approved):
Elegant solution 👍🏻

jsravn (changes requested):
I think this can be even more generic than "external" since it will work with any setup that has static domains for feed. So I suggest we rename it to "static" and update the README as well.

feed-ingress/cmd/external.go
```diff
@@ -0,0 +1,46 @@
...
+
+	externalCmd.Flags().StringVar(&externalHostname, "external-hostname", "",
+		"Hostname for external ingress")
+	externalCmd.Flags().StringVar(&internalHostName, "internal-hostname", "",
+		"Hostname for internal ingress")
+
+}
+
+func empty(kubernetesClient k8s.Client, updaters []controller.Updater) ([]controller.Updater, error) {
+
```
jsravn, line 34: remove whitespace

chbatey:
Will re-name to static and update readme 👍

jsravn (approved):

PR #243, opened 2022-02-01
Add a readiness endpoint to feed.

aecay:
The readiness endpoint calls down to a readiness method on the
controller, just as the healthcheck endpoint does.  The readiness
endpoint is implemented as a loop over all the controllerʼs owned
objects to poll them for readiness.

Most object types implement readiness by just returning the result of
their healthcheck.  The ALB and ELB types keep track of whether they have been registered with AWS, and return that status as their readiness.  The the NLB object type is where interesting things happen.  The AWS docs
call out a situation where it can take “a few minutes” for
healthchecking of an instance registered with a target group to begin.
During this interval, k8s should consider that replica “healthy” (itʼs
registered with the NLB successfully), but not “ready” (itʼs not
receiving traffic).  The distinction matters when doing rolling
deployments.  K8s wonʼt consider a new replica to be fully deployed (and move on with the deployment)
until itʼs ready.  And we donʼt want to treat feed as fully deployed
until it begins receiving traffic from the NLB target group (because we don't want to move on with deploying more replicas until the ones we have deployed are receiving traffic).

This situation has caused several production wobbles for us, where the
premature marking of feed pods as ready leads a rolling deployment to
complete too fast.  In the pessimistic case, we ended up with a brief
interval where there were no live replicas on the NLB at all (because
the old replicas had been torn down, but none of the new ones were out
of their NLB initial healthcheck delay yet).

The readiness logic for NLBs is so far implemented as a simple 5 minute
delay from initial NLB registration to the time when the NLB
controller (and by extension the entire feed process) will be marked as
“ready”.  In the cases weʼve seen, 5 minutes has been a sufficient
delay.  Thereʼs a possible better implementation, whereby we actually
observe whether weʼve received a healthcheck from NLB, and mark
ourselves as ready at the time when that occurs (which could be <5
minutes, leading to more responsive deployments, or could in theory be
\>5 minutes, giving us a safety net if the operational behavior of AWS
NLB changes).  However, thatʼs difficult to implement.  K8s and AWS
healthchecks arrive on the same endpoint.  The NLB controller has no
visibility of the request that is triggering a call to its health
method.  So the mooted implementation would require extensive
refactoring to be able to detect the "healthcheck from NLB" condition
specifically.

This implementation is good enough for now, and the better implementation
can be slotted in if/when it becomes a priority.

Having this implemented in feed itself allows us to avoid the need to
implement workarounds in the deployment yaml of feed in
kubernetes (replicating this implementation with
`initialReadinessDelay`, and not being forward-compatible with any
improvements).  It also moves the problem out of the realm of shadow
knowledge and into the open source implementation.

### DIFF
--- a/alb/alb.go
+++ b/alb/alb.go
@@ -125,6 +125,9 @@ func (a *alb) Health() error {
 }
 
 func (a *alb) Readiness() error {
+	if !a.readyForHealthCheck.Get() {
+		return errors.New("ALB registration not attempted yet")
+	}
 	return a.Health()
 }
 
--- a/alb/alb.go
+++ b/alb/alb.go
@@ -132,7 +132,7 @@ func (a *alb) Readiness() error {
 }
 
 func (a *alb) String() string {
-	return "ELB frontend"
+	return "ALB frontend"
 }
 
 func (a *alb) attachToFrontEnds() error {
--- a/alb/alb.go
+++ b/alb/alb.go
@@ -124,6 +124,10 @@ func (a *alb) Health() error {
 	return fmt.Errorf("have not attached to all frontends %v yet", a.targetGroupNames)
 }
 
+func (a *alb) Readiness() error {
+	return a.Health()
+}
+
 func (a *alb) String() string {
 	return "ELB frontend"
 }
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -57,7 +57,7 @@ type Controller interface {
 	Stop() error
 	// Health returns nil for a healthy controller, an error for unhealthy.
 	Health() error
-	// Readiness returns nil for a ready controller, an error for unhealthy.
+	// Readiness returns nil for a ready controller, an error for unready.
 	Readiness() error
 }
 
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -55,8 +55,10 @@ type Controller interface {
 	Start() error
 	// Stop the controller, blocking until it stops or an error occurs.
 	Stop() error
-	// Healthy returns true for a healthy controller, false for unhealthy.
+	// Health returns nil for a healthy controller, an error for unhealthy.
 	Health() error
+	// Readiness returns nil for a ready controller, an error for unhealthy.
+	Readiness() error
 }
 
 type controller struct {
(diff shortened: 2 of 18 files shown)


alb/alb.go
```diff
@@ -124,6 +124,13 @@ func (a *alb) Health() error {
 	return fmt.Errorf("have not attached to all frontends %v yet", a.targetGroupNames)
 }
 
+func (a *alb) Readiness() error {
+	if !a.readyForHealthCheck.Get() {
```
alangibson01, line 128: If I understand correctly, the difference in the way that the `Health` and `Readiness` functions are implemented could mean that Feed reports as ready but not healthy. Couldn't that cause some problems?

alb/alb.go
```diff
@@ -124,6 +124,13 @@ func (a *alb) Health() error {
...
+		return errors.New("ALB registration not attempted yet")
+	}
+	return a.Health()
+}
+
 func (a *alb) String() string {
 	return "ELB frontend"
```
alangibson01, line 135:
```suggestion
	return "ALB frontend"
```

Not related to your changes obviously, but while we're in this file we might as well correct this typo.

controller/controller.go
```diff
@@ -55,8 +55,10 @@ type Controller interface {
 	Start() error
 	// Stop the controller, blocking until it stops or an error occurs.
 	Stop() error
-	// Healthy returns true for a healthy controller, false for unhealthy.
+	// Health returns nil for a healthy controller, an error for unhealthy.
 	Health() error
+	// Readiness returns nil for a ready controller, an error for unhealthy.
```
alangibson01, line 60:
```suggestion
	// Readiness returns nil for a ready controller, an error for unready.
```

alangibson01 (changes requested):
Just a few comments and questions.

alb/alb.go
aecay: 1. I don't think it would cause problems
2. I don't even think it could happen.  If readyForHealthcheck == false, then it will report healthy but not ready.  If readyForHealthcheck == true, then ready will report exactly what healthy reports/

alangibson01 (approved):

PR #244, opened 2022-02-02
Upgrade vts module and provide support for request buckets

alangibson01:
Expose the nginx VTS module's `vhost_traffic_status_histogram_buckets` configuration property, allowing collection of latency information on requests flowing through feed.


aecay (approved):
lgtm - up to you whether you want to wait for a review from one of the old hands more familiar with feed's architecture, but to my eyes it's a simple straightforward change (so I would not say that another review is strictly needed).

PR #246, opened 2022-04-06
feat(nginx): allow nginx to proxy WebSocket headers

pyoio:
Allows feed to support proxying WebSocket requests if explicitly enabled.

If enabled via the `--nginx-allow-websocket-upgrade` flag, the nginx configuration will be generated with the following block:

```
    # Support WebSocket upgrade, allow keepalive
    # Upgrade logic from http://nginx.org/en/docs/http/websocket.html
    map $http_upgrade $connection_upgrade {
        default upgrade;
        ''      '';
    }

    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection $connection_upgrade;
```

Which will set the `Connection` header to `upgrade` IF the `Upgrade` header is present, otherwise the `proxy_set_header Connection` will be `""` which does not forward, the same as existing feed behaviour.

If the flag is omitted, the nginx config will generate with:

```
    # Enable keepalive to backend.
    proxy_http_version 1.1;
    proxy_set_header Connection "";
```

as it previously did.

### DIFF
--- a/nginx/nginx_test.go
+++ b/nginx/nginx_test.go
@@ -313,6 +313,9 @@ func TestNginxConfig(t *testing.T) {
 	noVhostStatsRequestBucketsConf := defaultConf
 	noVhostStatsRequestBucketsConf.VhostStatsRequestBuckets = nil
 
+	websocketProxyConf := defaultConf
+	websocketProxyConf.AllowWebsocketUpgrade = true
+
 	var tests = []struct {
 		name             string
 		conf             Conf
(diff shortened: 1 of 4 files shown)


nginx/nginx_test.go
```diff
@@ -523,6 +526,21 @@ func TestNginxConfig(t *testing.T) {
...
+				"# Upgrade logic from http://nginx.org/en/docs/http/websocket.html",
+				"map $http_upgrade $connection_upgrade {",
+				"    default upgrade;",
+				"    ''      '';",
+				"}",
+				"proxy_http_version 1.1;",
+				"proxy_set_header Upgrade $http_upgrade;",
+				"proxy_set_header Connection $connection_upgrade;",
+			},
+		},
```
pyoio, line 543: This may be too verbose? Happy to shorten if preferred

pyoio:
Meant to ping @sky-uk/core-networking en masse to discuss, didn't realise it would randomly select @aecay 😬

pyoio:
The use case for us is to enable WebSocket connections in our mgmt cluster for mgmt tools (e.g., grafana, tekton ui) -- no plans to enable it in other clusters

aecay (changes requested):
I'm not convinced of the value of this PR.  To summarize what I said on slack:

- There are unconsidered performance implications to this change
- As written, it allows websockets globally for all ingresses on the platform rather than per-ingress
- We don't understand how we could deploy it successfully (haproxy, graceful termination support)
- There's no genuine requirement to support this (and by extension I don't think we have the ability to ever support it)

So, this is a NACK from me.  We're continuing to discuss but I want to be clear about what I see the obstacles as being.

pyoio:
Short term we're going to flip our mgmt tools w/ websockets over to using cloud native LBs which support WS, the support + HAProxy aspects make this something that's more appropriately considered in next-gen ingress work when that comes IMO, cheers @aecay!

PR #249, opened 2022-07-29
update ingress api to networking/v1

aecay:
This brings feed in line with newer versions of kubernetes which (optionally, and then obligatorily) use this API in preverence to the extensions/v1beta1 one.


pyoio (approved):
Nice! Much more pleasant than our attempt at having v1beta1 and networking v1 co-existing in the same codebase 😂

howardburgess (approved):

aecay:
> Nice! Much more pleasant than our attempt at having v1beta1 and networking v1 co-existing in the same codebase joy

I...did not know you had tried that and I'm glad I didn't, otherwise I probably would have run screaming :sweat_smile:

howardburgess (approved):

PR #252, opened 2023-04-18
exit-when-nginx-reload-sig-failed

mohamediag:
this to ensure to exit when nginx reload fails due to master process being killed
for https://github.com/sky-uk/core-networking/issues/3161

### DIFF
--- a/nginx/nginx.go
+++ b/nginx/nginx.go
@@ -95,7 +95,10 @@ func (n *nginxUpdater) signalIfRequired() {
 	if n.updateRequired.Get() {
 		log.Info("Signalling Nginx to reload configuration")
 		incrementReloadMetric()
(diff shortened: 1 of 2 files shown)


nginx/nginx.go
```diff
@@ -95,7 +95,10 @@ func (n *nginxUpdater) signalIfRequired() {
 	if n.updateRequired.Get() {
 		log.Info("Signalling Nginx to reload configuration")
 		incrementReloadMetric()
```
chaosbox, line 97: can we move this after we have sent the signal pls

chaosbox: *both lines

mohamediag: done

chaosbox (approved):

PR #255, opened 2023-07-21
Truncate comment to fit under the line limit for nginx config

barisi:
When a large number of ingresses are created, we get into a situation where the line length on the ingresses comment exceeds the maximum allowed for an nginx config line and this stops nginx from starting up.

### DIFF
--- a/nginx/nginx.tmpl
+++ b/nginx/nginx.tmpl
@@ -160,9 +160,7 @@ http {
 {{ end }}
 
 {{- range $entry := .Servers }}
-    {{ $length := len $entry.Name }}
-    {{- if lt $length 4000 }}# ingress: {{ $entry.Name }}{{ end }}
-    {{- if gt $length 4000 }}# ingress: {{ printf "%.4000s" $entry.Name }}{{ end }}
+    # ingress: {{ printf "%.4000s" $entry.Name }}
   {{- range $portConf := $IngressPorts }}
     server {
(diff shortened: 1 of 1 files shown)


chaosbox:
```
/ $ nginx -T
2023/07/21 09:38:55 [emerg] 49#49: too long parameter "# ingress:..." started in /nginx/nginx.conf:65493
nginx: [emerg] too long parameter "# ingress:..." started in /nginx/nginx.conf:65493
nginx: configuration file /nginx/nginx.conf test failed
```

One of the tenant's namespace had too many ingresses that caused the commented line to exceed the limit.

Should we remove it altogether ? or multiline comment? Truncating makes the comment unusable, I think.

chaosbox:
If we are truncating, then I think we should have comment line saying "following line is truncated to 4000 characters due to limits"?

```
{{ $strLen := len $entry.Name }} {{ if gt $strLen 4000 }}
# following line is truncated to 4000 characters due to limits
{{ end }}
```

barisi:
> If we are truncating, then I think we should have comment line saying "following line is truncated to 4000 characters due to limits"?
> 
> ```
> {{ $strLen := len $entry.Name }} {{ if gt $strLen 4000 }}
> # following line is truncated to 4000 characters due to limits
> {{ end }}
> ```

Updated the PR with a comment to indicate truncated lines.

chaosbox (approved):
lgtm

